PR #160, opened 2022-02-08
Geometry nodes support

seandillon92:
Support for sending instance info from geometry nodes


unity-cla-assistant:
[CLA assistant check](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/MeshSyncDCCPlugins?pullRequest=160) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/MeshSyncDCCPlugins?pullRequest=160) before we can accept your contribution.<br/><hr/>**Sean Dillon** seems not to be a GitHub user. You need a GitHub account to be able to sign the CLA. If you have already a GitHub account, please [add the email address used for this commit to your account](https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/#commits-are-not-linked-to-any-user).<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.cds.internal.unity3d.com/check/Unity-Technologies/MeshSyncDCCPlugins?pullRequest=160) it.</sub>

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
```diff
@@ -1412,10 +1436,89 @@ bool msblenContext::sendObjects(MeshSyncClient::ObjectScope scope, bool dirty_al
...
+            continue;
+        
+        // If the object data is null, skip
+        auto data = (ID*)object->data;
+        if (data == nullptr)
+            continue;
+
+        auto name = object->id.name;
+        if (name != nullptr) {
+            if (object_instances.find(name) == object_instances.end()) {
```
SonnyCampbellUnity: is the result of `object_instances.find(name)` the as the `instance` retrieved using `object_instances_get` above?

seandillon92: No, object_instances is the extracted set that maps an object name to an array of matrices. Object_instances_get is a method to access the actual instance from the iterator

Plugins~/Src/MeshSyncClientBlender/msblenContext.h
```diff
@@ -45,6 +45,8 @@ class msblenContext {
     bool sendAnimations(MeshSyncClient::ObjectScope scope);
     bool ExportCache(const std::string& path, const BlenderCacheSettings& cache_settings);
 
+    bool extractObjectInstances();
```
sindharta: we might need to put
```
#f BLENDER_VERSION >= 300
```
to enable these functions

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
```diff
@@ -1412,10 +1442,103 @@ bool msblenContext::sendObjects(MeshSyncClient::ObjectScope scope, bool dirty_al
         eraseStaleObjects();
     }
 
+    //TODO-Sean Dillon: cleanup object instances array
```
sindharta: SceneExporter has the following callbacks, which we can use 

```
std::function<void()> on_prepare, on_success, on_error, on_complete;
```

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
seandillon92: I will try to address this by creating an object that abstracts the object instances API. Defining it in the Context class requires me to add object instances in the naming

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
seandillon92: That should work perfectly, thank you!

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
```diff
@@ -704,6 +709,12 @@ ms::MeshPtr msblenContext::exportMesh(const Object *src)
     }
 
     if (data) {
+
+#if BLENDER_VERSION >= 300
+        // Inject instance info on the mesh
+        m_geometry_nodes.onMeshExport(src, dst);
+#endif
```
seandillon92: This will be redundant, we shouldn't inject code like this

Plugins~/Src/MeshSyncClientBlender/msblenGeometryNodes.h
```diff
@@ -0,0 +1,31 @@
...
+#include "pch.h"
+#include "MeshUtils/muMath.h"
+#include "DNA_object_types.h"
+#include "MeshSync/SceneGraph/msMesh.h"
+#include <DNA_node_types.h>
+
+namespace blender {
+
+	class msblenGeometryNodes
+	{
```
seandillon92: This class will become an instances manager, similar to entities manager etc. The idea is to decouple instancing data from objects, as maintaining dirty states will become more convoluted and it would be better we only send across instances, without the mesh info

seandillon92 (commented):
Hey @sindharta here is a self-review with some changes I would like to make in this PR. On a higher level, I noticed that while the VariantData works well for manual sync, auto sync requires some state management around what is dirty or not. 

A quick and dirty solution to the latest commit is to change the Transform checksum to take into account the variant data checksums. 

A cleaner solution would be to create a manager class for instances, similar to the entityManager. That would allow us to create our own serializable objects for instances and additional info if required. It would also decouple the instances info from meshes, so we should be able to send scene updates for instances without re-sending the mesh data.

Let me know what you think about the plan, any feedback would be appreciated!

Plugins~/Src/MeshSyncClientBlender/msblenGeometryNodes.h
sindharta: The idea of decoupling is good. 
But obviously we still need the link to the original object, and we also need to handle outlier cases, for example if for some reason, the user happens to delete the original object on the Unity side.

sindharta: And FYI, I believe only the required mesh info is sent to Unity at the moment.
For example, if we are not updating (dirtying) the transform, then I think the transform data will be skipped.

sindharta:
A general question. Without this PR on the Meshsync side: [549](https://github.com/unity3d-jp/MeshSync/pull/540/files), will Unity just simply ignore the instance info ?
Might be worth testing to ensure it doesn't crash.

sindharta:
Another idea is, can we make a certain user interaction to set a flag for sending geometry node variant data, instead of having to calculate the checksum of all variant data.

seandillon92:
> A general question. Without this PR on the Meshsync side: [549](https://github.com/unity3d-jp/MeshSync/pull/540/files), will Unity just simply ignore the instance info ? Might be worth testing to ensure it doesn't crash.

I think it will be safe. Unity will simply not check for the variant data. I will  definitely test it before merging in any case.

seandillon92:
> Another idea is, can we make a certain user interaction to set a flag for sending geometry node variant data, instead of having to calculate the checksum of all variant data.

That's a good point. At the moment I have started moving away from the VariantData approach but if we were to merge the variant data solution, I will make sure to implement this

seandillon92 (commented):
Implemented the Instance manager class and updated the scene exported to use it.  Need to validate thread safety.

Plugins~/Src/MeshSyncClientBlender/msblenGeometryNodes.h
seandillon92: For linking the objects, I use a path in my instance info class that matches the entity path, so the matching works at a record level on the Unity Editor. For deletions, I extended the delete message to take instance info identifiers and the instance renderer listens to instance info deletions to update its own list of meshes.

Plugins~/Src/MeshSyncClientBlender/msblenContext.h
```diff
@@ -121,14 +133,15 @@ class msblenContext {
...
-    ms::MeshPtr exportMesh(const Object *obj);
+    ms::TransformPtr exportObject(const Object* obj, bool parent, ms::TransformManager& manager, bool tip = true);
+    ms::TransformPtr exportTransform(const Object *obj, ms::TransformManager& manager);
+    ms::TransformPtr exportPose(const Object *armature, bPoseChannel *obj, ms::TransformManager& manager);
+    ms::TransformPtr exportArmature(const Object *obj, ms::TransformManager& manager);
+    ms::TransformPtr exportReference(Object *obj, const DupliGroupContext& ctx, ms::TransformManager& manager);
+    ms::TransformPtr exportDupliGroup(const Object *obj, const DupliGroupContext& ctx, ms::TransformManager& manager);
+    ms::CameraPtr exportCamera(const Object *obj, ms::TransformManager& manager);
+    ms::LightPtr exportLight(const Object *obj, ms::TransformManager& manager);
+    ms::MeshPtr exportMesh(const Object *obj, ms::TransformManager& manager);
```
seandillon92: Hey @sindharta, for reusing the exporting code for objects that are not exported as entities but they have to be exported as inputs to geometry nodes instancing, I had to refactor a lot of these callbacks to take a Transform manager as a dependency. I noticed that there are a lot of other data structures associated with the entity_manager instance, for example the m_obj_records, the m_pending etc, so I plan to refactor the code in this way:

Create a new class named blenContextState that keeps all these objects as members. I will then have two instances of this class, one for the entity_manager and one for the instances_manager. This will allow me to pass the state object around and avoid having a data clamp. In the future, I was thinking that a lot of the exportXXX functions could be moved to the state class or to another object. That way it should be easier to decouple and test some of these functions and the code would follow the Open Closed and Single Responsibility principles more. Do you agree about this change?

Plugins~/Src/MeshSyncClientBlender/msblenUtils.cpp
```diff
@@ -4,121 +4,156 @@
 #include "msblenUtils.h"
 
 namespace bl = blender;
+namespace msblenUtils {
```
sindharta, line 7: Hm, the changes are a little bit hard to read when a tab or spaces are added in every line.

Perhaps we can just:
1. add `msblenUtils` namespace 
2. leave the other code as is 
3. add spaces in another PR which just focuses on modifying code convention ?

My personal choice though, is to leave the code starting from the leftmost in this case, to save space.

Plugins~/Src/MeshSyncClientBlender/msblenGeometryNodes.h
```diff
@@ -0,0 +1,68 @@
...
+#include "MeshSync/SceneGraph/msMesh.h"
+#include <DNA_node_types.h>
+#include "MeshSyncClient/msInstancesManager.h"
+#include "MeshUtils/muMath.h"
+#include <DNA_mesh_types.h>
+
+namespace blender {
+
+#if BLENDER_VERSION >= 300
+	class GeometryNodesUtils
```
sindharta: Is this using a tab with 8 spaces ?
Just wanted to make sure if this is intentional.
We'll also probably want to agree on the tab/spaces convention.

Plugins~/Src/MeshSyncClient/Include/MeshSyncClient/msInstancesManager.h
```diff
@@ -0,0 +1,54 @@
...
+namespace ms {
+
+    class InstancesManager : public TransformManager
+    {
+    public:
+        std::vector<TransformPtr> getDirtyMeshes();
+        std::vector<InstanceInfoPtr> getDirtyInstances();
+        std::vector<Identifier>& getDeleted();
+        void clearDirtyFlags();
+        void add(InstanceInfoPtr instanceInfo);
```
sindharta: making it clear that we are overriding virtual functions would be good for code maintenance

Plugins~/Src/MeshSyncClient/Include/MeshSyncClient/msEntityManager.h
```diff
@@ -1,13 +1,14 @@
...
 msDeclClassPtr(Transform)
 
 #ifndef msRuntime
 namespace ms {
 
-class EntityManager
+class EntityManager : public TransformManager
 {
 public:
     EntityManager();
```
sindharta, line 29: making it clear that we are overriding virtual functions would be good for code maintenance

Plugins~/Src/MeshSyncClient/msInstancesManager.cpp
```diff
@@ -0,0 +1,112 @@
...
+        rec.entity = entity;
+    }
+
+    
+
+    void InstancesManager::add(InstanceInfoPtr info)
+    {
+        auto& rec = lockAndGet(info->path);
+
+        //TODO implement hash for InstanceInfo and check
```
sindharta: Is there a performance issue if we are not implementing this ?

Plugins~/Src/MeshSyncClient/msInstancesManager.cpp
```diff
@@ -0,0 +1,112 @@
...
+            if (!it->second.updated) {
+                m_deleted.push_back(it->second.instances->getIdentifier());
+                m_records.erase(it++);
+            }
+            else
+                ++it;
+        }
+    }
+
+    void InstancesManager::setAlwaysMarkDirty(bool alwaysDirty) {
```
sindharta: Same with `EntityManager::setAlwaysMarkDirty`

Maybe we can avoid using virtual here ?

Plugins~/Src/MeshSyncClient/msInstancesManager.cpp
```diff
@@ -0,0 +1,112 @@
...
+        m_always_mark_dirty = alwaysDirty;
+    }
+
+    InstancesManager::Record& InstancesManager::lockAndGet(const std::string& path)
```
sindharta: Same with `EntityManager::lockAndGet`

Maybe we can avoid using virtual here ?

Plugins~/Src/MeshSyncClientBlender/BlenderPyObjects/BlenderPyDepsgraphObjectInstance.h
```diff
@@ -0,0 +1,22 @@
...
+#include "MeshUtils/muMath.h"
+
+namespace blender
+{
+    class BlenderPyDepsgraphInstance
+    {
+        public:
+            BlenderPyDepsgraphInstance(PointerRNA& instance) : m_instance(instance) {}
+
+            Object* instance_object();
```
sindharta, line 13: We should probably agree on the code convention for functions as well.

Plugins~/Src/MeshSyncClientBlender/BlenderPyObjects/BlenderPyDepsgraphObjectInstance.cpp
```diff
@@ -0,0 +1,53 @@
...
+namespace blender {
+
+    PropertyRNA* BlenderPyDepsgraphObjectInstance_instance_object = nullptr;
+    PropertyRNA* BlenderPyDepsgraphObjectInstance_is_instance = nullptr;
+    PropertyRNA* BlenderPyDepsgraphObjectInstance_world_matrix = nullptr;
+    PropertyRNA* BlenderPyDepsgraphObjectInstance_parent = nullptr;
+    PropertyRNA* BlenderPyDepsgraphObjectInstance_object = nullptr;
+
+    Object* BlenderPyDepsgraphInstance::instance_object() {
+        auto objectInstanceProp = (PointerPropertyRNA*)BlenderPyDepsgraphObjectInstance_instance_object;
```
sindharta, line 14: we should probably agree on the code convention for using `auto/var` as well.

Plugins~/CMakeLists.txt
```diff
@@ -53,7 +53,11 @@ include(UnityPackage)
 setup_unity_package_version()
```
sindharta, line 53: I have this error when trying to build a plugin using this branch

`Cannot open include file: 'MeshSync/SceneGraph/msInstanceInfo.h'`

Plugins~/Src/MeshSyncClientBlender/msblenContextIntermediatePathProvider.h
```diff
@@ -0,0 +1,8 @@
+#pragma once
+#include <msblenContextPathProvider.h>
+
+class msblenContextIntermediatePathProvider: public msblenContextPathProvider {
+public: 
+	std::string get_path(const Object* obj);
```
sindharta: override ?

Plugins~/Src/MeshSyncClientBlender/msblenContextDefaultPathProvider.h
```diff
@@ -0,0 +1,13 @@
+#pragma once
+#include <string>
+#include <pch.h>
+#include <msblenContextPathProvider.h>
+#include <msblenUtils.h>
+
+class msblenContextDefaultPathProvider : public msblenContextPathProvider {
+public:
+	std::string get_path(const Object* obj);
```
sindharta: override ?

Plugins~/Src/MeshSyncClientBlender/msblenGeometryNodes.cpp
```diff
@@ -0,0 +1,167 @@
...
+
+                // Check if the data names also match
+                auto recDataId = (ID*)rec->second->object_copy.data;
+                auto sceneDataId = (ID*)obj->data;
+
+                if (strcmp(sceneDataId->name + 2, recDataId->name + 2) != 0)
+                    continue;
+
+                handler(obj, rec->second->parent, std::move(rec->second->matrices), true);
+                rec->second->handled = true;
```
sindharta: this `handled` variable seems to be used only here, so I am wondering if we should store it inside the `Record` data structure ?

For example, calling this `each_instanced_object()` function twice will probably produce different results, and I am wondering if this is intentional ?

Plugins~/Src/MeshSyncClientBlender/msblenContextState.cpp
```diff
@@ -0,0 +1,83 @@
...
+    // care children
+    if (children) {
+        each_child(obj, [&](Object* child) {
+            touchRecord(paths, child, base_path, true);
+            });
+    }
+
+    // trace dupli group
+    if (Collection* group = get_instance_collection(obj)) {
+        const std::string group_path = path + '/' + (group->id.name + 2);
```
sindharta, line 75: what is 2 ?

Plugins~/Src/MeshSyncClientBlender/msblenContextPathProvider.h
```diff
@@ -0,0 +1,9 @@
+#pragma once
+
+#include <string>
+
+class msblenContextPathProvider {
+public:
+	virtual std::string get_path(const Object* obj) = 0;
+	virtual std::string get_path(const Object* arm, const Bone* obj) = 0;
```
sindharta: it seems that we can combine these two functions into this ?

`virtual std::string get_path(const Object* obj, const Bone* obj = nullptr) = 0;`

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
```diff
@@ -56,34 +62,44 @@ void msblenContext::NodeRecord::recordAnimation(msblenContext *_this) const {
...
 //----------------------------------------------------------------------------------------------------------------------
 
-msblenContext& msblenContext::getInstance() {
-    static msblenContext s_instance;
-    return s_instance;
+static msblenContext* s_instance = nullptr;
+
+msblenContext& msblenContext::getInstance()
+{
+    if (s_instance == nullptr) {
```
sindharta: Why do we need to change the singleton creation this way ?
The previous way is using this Singleton design pattern, and should be safe/robust.
https://stackoverflow.com/questions/1008019/c-singleton-design-pattern

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
```diff
@@ -56,34 +62,44 @@ void msblenContext::NodeRecord::recordAnimation(msblenContext *_this) const {
 }
 
 
-void msblenContext::ObjectRecord::clearState()
```
sindharta, line 59: Was this moved to `msblenContextState::ObjectRecord` ?
If so, I think it is better if we refactor the code first and create another PR before adding new functionality.

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
```diff
@@ -56,34 +62,44 @@ void msblenContext::NodeRecord::recordAnimation(msblenContext *_this) const {
...
+    m_instances_state->clear();
+
+    delete s_instance;
 }
 
 
 
 msblenContext::msblenContext()
 {
+    m_entities_state = new msblenContextState(m_entity_manager);
```
sindharta: Are we sure that we want to manage new/destroy manually rather than using smart pointers ?

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
```diff
@@ -290,11 +306,11 @@ static void extract_bone_trs(const mu::float4x4& mat, mu::float3& t, mu::quatf&
     s = mu::swap_yz(s);
 }
 
-void msblenContext::extractTransformData(const Object *obj,
+void msblenContext::extractTransformData(BlenderSyncSettings& settings, const Object *obj,
     mu::float3& t, mu::quatf& r, mu::float3& s, ms::VisibilityFlags& vis,
     mu::float4x4 *dst_world, mu::float4x4 *dst_local)
 {
-    vis = { true, visible_in_render(obj), visible_in_viewport(obj) };
+    vis = { visible_in_collection(obj), visible_in_render(obj), visible_in_viewport(obj) };
```
sindharta: Is this a change addressing this issue ?
https://github.com/Unity-Technologies/MeshSyncDCCPlugins/issues/165

If so, could you please make a separate PR for this ?

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
```diff
@@ -290,11 +306,11 @@ static void extract_bone_trs(const mu::float4x4& mat, mu::float3& t, mu::quatf&
     s = mu::swap_yz(s);
 }
 
-void msblenContext::extractTransformData(const Object *obj,
+void msblenContext::extractTransformData(BlenderSyncSettings& settings, const Object *obj,
```
sindharta: Do we ever pass variables other than `m_settings` here ?

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
```diff
@@ -481,85 +502,87 @@ ms::TransformPtr msblenContext::exportObject(const Object *obj, bool parent, boo
...
 
-            exportDupliGroup(obj, ctx);
+            exportDupliGroup(state, paths, settings, obj, ctx);
         }
     }
     return rec.dst;
 }
 
-ms::TransformPtr msblenContext::exportTransform(const Object *src)
+ms::TransformPtr msblenContext::exportTransform(msblenContextState& state, msblenContextPathProvider& paths, BlenderSyncSettings& settings, const Object *src)
```
sindharta, line 501: It's probably better if we create another PR to do this refactoring (add three parameters in the function), as this refactoring itself contains quite a lot of modification

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
```diff
@@ -481,85 +502,87 @@ ms::TransformPtr msblenContext::exportObject(const Object *obj, bool parent, boo
...
     std::string path = ctx.dst->path + local_path;
 
     ms::TransformPtr dst;
     auto assign_base_params = [&]() {
-        extractTransformData(src, *dst);
+        extractTransformData(settings, src, *dst);
         dst->path = path;
-        // todo:
-        dst->visibility = {};
+        dst->visibility = {visible_in_collection(src), visible_in_render(ctx.group_host), visible_in_viewport(ctx.group_host)};
```
sindharta: This change is probably addressing https://github.com/Unity-Technologies/MeshSyncDCCPlugins/issues/165 ?
Might be good to put it in a different PR.

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
```diff
@@ -603,52 +626,52 @@ ms::TransformPtr msblenContext::exportDupliGroup(const Object *src, const DupliG
 
     std::shared_ptr<ms::Transform> dst = ms::Transform::create();
     dst->path = path;
-    dst->visibility = { true, visible_in_render(ctx.group_host), visible_in_viewport(ctx.group_host) };
+    dst->visibility = { visible_in_collection(src), visible_in_render(ctx.group_host), visible_in_viewport(ctx.group_host) };
```
sindharta: This change is probably addressing https://github.com/Unity-Technologies/MeshSyncDCCPlugins/issues/165 ?
Might be good to put it in a different PR.

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
```diff
@@ -603,52 +626,52 @@ ms::TransformPtr msblenContext::exportDupliGroup(const Object *src, const DupliG
...
     ctx2.group_host = src;
     ctx2.dst = dst;
     auto gobjects = bl::list_range((CollectionObject*)group->gobject.first);
     for (auto go : gobjects) {
         auto obj = go->ob;
-        if (auto t = exportObject(obj, true, false)) {
+        if (auto t = exportObject(state, paths, settings, obj, true, false)) {
             const bool non_lib = obj->id.lib == nullptr;
-            t->visibility = { true, non_lib, non_lib };
+            t->visibility = { visible_in_collection(obj), non_lib, non_lib };
```
sindharta, line 635: This change is probably addressing https://github.com/Unity-Technologies/MeshSyncDCCPlugins/issues/165 ?
Might be good to put it in a different PR.

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
```diff
@@ -1086,75 +1111,19 @@ void msblenContext::doExtractEditMeshData(ms::Mesh& dst, const Object *obj, Mesh
...
-ms::TransformPtr msblenContext::findBone(Object *armature, Bone *bone)
+ms::TransformPtr msblenContext::findBone(msblenContextState& state, Object *armature, Bone *bone)
 {
-    std::map<struct Bone*, std::shared_ptr<ms::Transform>>::iterator it = m_bones.find(bone);
-    return it != m_bones.end() ? it->second : nullptr;
+    std::map<struct Bone*, std::shared_ptr<ms::Transform>>::iterator it = state.bones.find(bone);
+    return it != state.bones.end() ? it->second : nullptr;
 }
 
-msblenContext::ObjectRecord& msblenContext::touchRecord(const Object *obj, const std::string& base_path, bool children)
```
sindharta, line 1094: This has been moved to `msblenContextState::touchRecord`, I guess ?
Might be good to put the move refactoring in a different PR.

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
```diff
@@ -1635,6 +1634,130 @@ void msblenContext::WaitAndKickAsyncExport()
...
+    m_geometryNodeUtils.each_instanced_object([this, &scene_objects](Object* instanced, Object* parent, SharedVector<mu::float4x4> matrices, bool fromFile){
+        // If the instanced object is not present in the file
+        if (!fromFile) {
+            return exportInstancesFromTree(instanced, parent, std::move(matrices));
+        }
+
+        // check if the object has been already exported as part of the scene
+        auto scene_object = scene_objects.find(instanced);
+        if (scene_object == scene_objects.end()) {
+            return exportInstacesFromFile(instanced, parent, std::move(matrices));
```
sindharta: typo.
`exportInstacesFromFile` vs
`exportInstancesFromFile`

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
```diff
@@ -1635,6 +1634,130 @@ void msblenContext::WaitAndKickAsyncExport()
...
+        }
+        else {
+            return exportInstacesFromScene(instanced, parent, std::move(matrices));
```
sindharta: similar typo

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
```diff
@@ -1635,6 +1634,130 @@ void msblenContext::WaitAndKickAsyncExport()
...
+void msblenContext::onDepsgraphUpdatedPost(Depsgraph* graph)
+{
+#if BLENDER_VERSION >= 300
+    m_geometryNodeUtils.setInstancesDirty(true);
+#endif
+
+}
+
+/// Geometry Nodes Blender Context Functionality ///
+#if BLENDER_VERSION >= 300
```
sindharta: Maybe it's better to put the geometry node functionality in a different file

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
```diff
@@ -1635,6 +1634,130 @@ void msblenContext::WaitAndKickAsyncExport()
...
+    auto inverse = mu::invert(object_world_matrix);
+;
+    mu::parallel_for(0, mat.size(), 10, [this, &mat, &inverse](int i) 
+        {
+            mat[i] = m_geometryNodeUtils.blenderToUnityWorldMatrix(mat[i] * inverse);
+        });
+
+    exportInstanceInfo(*m_instances_state, m_default_paths, instancedObject, parent, std::move(mat));
+}
+void msblenContext::exportInstacesFromScene(Object* instancedObject, Object* parent, SharedVector<mu::float4x4> mat)
```
sindharta: Almost the same as `exportInstacesFromFile` ?
Maybe we can reduce code duplicity ?

Plugins~/Src/MeshSyncClient/Include/MeshSyncClient/msTransformManager.h
```diff
@@ -0,0 +1,37 @@
+#pragma once
+
+#include "pch.h"
+#include "MeshSync/MeshSync.h"
+
+namespace ms{
+	class TransformManager {
```
John-Hackett: Can you add some documentation for this class?

Plugins~/Src/MeshSyncClient/Include/MeshSyncClient/msInstancesManager.h
```diff
@@ -0,0 +1,54 @@
...
+#include <unordered_map>
+#include "msTransformManager.h"
+
+#ifndef msRuntime
+
+msDeclClassPtr(InstanceInfo)
+
+namespace ms {
+
+    class InstancesManager : public TransformManager
```
John-Hackett: Can you add some documentation for this class?

Plugins~/Src/MeshSyncClient/Include/MeshSyncClient/msTransformManager.h
John-Hackett: Would be nice to have some for the functions as well

Plugins~/Src/MeshSyncClientBlender/BlenderPyObjects/BlenderPyCommon.h
```diff
@@ -2,6 +2,7 @@
 
 #include <RNA_define.h> //PropertyRNA
 #include "MeshSync/msFoundation.h" //msPacked
+#include "msblenMacros.h"
```
John-Hackett, line 6: Is this import needed?

Plugins~/Src/MeshSyncClientBlender/BlenderPyObjects/BlenderPyDepsgraph.h
```diff
@@ -0,0 +1,21 @@
...
+#include <BKE_context.h> //bContext
+#include "MeshUtils/muMath.h"
+
+namespace blender
+{
+
+    class BlenderPyDepsgraph
+    {
+    public:
+        BlenderPyDepsgraph(Depsgraph* depsgraph);
```
John-Hackett, line 18: Documentation needed

Plugins~/Src/MeshSyncClientBlender/msblenBinder.cpp
```diff
@@ -62,6 +64,19 @@ extern PropertyRNA* BlenderPyContext_blend_data;
 extern PropertyRNA* BlenderPyContext_scene;
 extern FunctionRNA* BlenderPyContext_evaluated_depsgraph_get;
 extern FunctionRNA* BlenderPyContext_depsgraph_update;
+extern PropertyRNA* BlenderPyContext_viewlayer;
```
John-Hackett: non blocking, this should be view_layer

Plugins~/Src/MeshSyncClientBlender/msblenContextState.cpp
seandillon92: this is moved code, need to check if I added this by accident. +2 would take the name without the first 2 characters. So instead of MEMyMesh it would just return MyMesh.

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
seandillon92: yes, @schinkowski has this on his branch, I backported it to mine so I can test more easily with the file he is working with. I can create a PR or maybe Christian could as its his solution. Will remove it from this one in any case, it was more of a convenience change, nothing blocking

Plugins~/CMakeLists.txt
seandillon92: is the MeshSync repo branch instance-rendering-support  or dev?

Plugins~/Src/MeshSyncClientBlender/BlenderPyObjects/BlenderPyCommon.h
seandillon92: Yes, it solves a compilation error with clang.

Plugins~/Src/MeshSyncClientBlender/msblenUtils.cpp
seandillon92: I applied the above should be easier to see the actual change now

Plugins~/Src/MeshSyncClientBlender/msblenGeometryNodes.h
seandillon92: its meant to be a 4 spaces tab. Yes I agree, would be good to decide on this

Plugins~/Src/MeshSyncClient/Include/MeshSyncClient/msEntityManager.h
seandillon92: added the override keyword

Plugins~/Src/MeshSyncClient/Include/MeshSyncClient/msInstancesManager.h
seandillon92: added the override keyword

Plugins~/Src/MeshSyncClient/msInstancesManager.cpp
seandillon92: I don't think so.  I think we might better without a hash check unless we have a good reason to do so.

Plugins~/Src/MeshSyncClient/msInstancesManager.cpp
seandillon92: I will change this but I think we should have another layer of abstraction that acts as an interface, if another implementation wants to change this behavior.

seandillon92: This is done now

Plugins~/Src/MeshSyncClient/msInstancesManager.cpp
seandillon92: This is done now

Plugins~/Src/MeshSyncClientBlender/BlenderPyObjects/BlenderPyDepsgraphObjectInstance.h
seandillon92: +1

Plugins~/Src/MeshSyncClientBlender/msblenContextPathProvider.h
seandillon92: Good point, done now

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
seandillon92: I need the c'tor where I can inject the manager dependencies for the state classes

seandillon92: I reverted the singleton pattern to how it was, found another way to solve my original problem

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
seandillon92: changed now to shared_ptr

Plugins~/Src/MeshSyncClientBlender/msblenBinder.cpp
seandillon92: Good catch! I ended up removing this code

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
seandillon92: Agreed, will try this

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
seandillon92: This is fixed now

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
seandillon92: Now fixed

Plugins~/Src/MeshSyncClientBlender/msblenGeometryNodes.cpp
seandillon92: I used the wrong naming convention for records there. I used to keep them as private members of the class and clearing the maps at the end of the operation. I realized later its simpler to keep them as local variables in the function and refactored my code but did not rename them. 

So the records do not persist between executions of the function, it will give the same results. Sorry for the confusion with the names

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
seandillon92: Agreed, will do so

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
seandillon92: Agreed, will do so

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
seandillon92: Yes, specifically for the bakeTransform option that is used in this function, I disable it when I export an object that is not part of the scene.

seandillon92: See msblenContext::exportInstancesFromFile for an example of this

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
seandillon92: Removed duplicate code

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
seandillon92: I moved these implementations to msblenContextGeometryNodes.cpp

Plugins~/Src/MeshSyncClient/Include/MeshSyncClient/msInstancesManager.h
seandillon92: Added XML comments on class and functions

Plugins~/Src/MeshSyncClientBlender/BlenderPyObjects/BlenderPyDepsgraph.h
seandillon92: Added XML comments on class and functions

Plugins~/Src/MeshSyncClient/Include/MeshSyncClient/msTransformManager.h
seandillon92: Added XML comments to class and interface

Plugins~/Src/MeshSyncClientBlender/msblenContextState.cpp
sindharta: If it's moved code, then I guess it's alright for now at least.

Plugins~/CMakeLists.txt
sindharta: Ah, it slipped my mind that we need to build this with the compatible branch in MeshSync as well.
Resolving

Plugins~/Src/MeshSyncClient/msInstancesManager.cpp
sindharta: Then, shouldn't we just remove the comment ?

Plugins~/Src/MeshSyncClientBlender/msblenGeometryNodeUtils.h
```diff
@@ -0,0 +1,68 @@
...
+#include "MeshSync/SceneGraph/msMesh.h"
+#include <DNA_node_types.h>
+#include "MeshSyncClient/msInstancesManager.h"
+#include "MeshUtils/muMath.h"
+#include <DNA_mesh_types.h>
+
+namespace blender {
+
+#if BLENDER_VERSION >= 300
+	class GeometryNodesUtils
```
sindharta: Seems to be using a tab with 8 spaces ?

Plugins~/Src/MeshSyncClientBlender/msblenContextGeometryNodes.cpp
```diff
@@ -0,0 +1,94 @@
...
+        auto inverse = mu::invert(world_matrix);
+
+        // check if the object has been already exported as part of the scene
+        auto scene_object = scene_objects.find(instanced);
+        if (scene_object == scene_objects.end()) {
+            exportObject(*m_instances_state, m_default_paths, settings, instanced, false);
+            return exportInstancesFromFile(instanced, parent, std::move(matrices), inverse);
+        }
+        else {
+            return exportInstancesFromFile(instanced, parent, std::move(matrices), inverse);
```
sindharta: same `return exportInstancesFromFile(instanced, parent, std::move(matrices), inverse);` in both paths ?

Plugins~/Src/MeshSyncClient/Include/MeshSyncClient/msTransformManager.h
```diff
@@ -0,0 +1,66 @@
...
+/// <typeparam name="T">The type of record that the manager uses</typeparam>
+template <class T>
+class TransformManager : public ITransformManager {
+public:
+
+    void setAlwaysMarkDirty(bool v) override {
+        m_always_mark_dirty = v;
+    }
+
+	virtual void add(TransformPtr transform) = 0;
```
sindharta: Some lines are using tabs ?
Not a big deal, but just can't help noticing them in github diff.

Plugins~/Src/MeshSyncClient/msInstancesManager.cpp
seandillon92: Its now removed

Plugins~/Src/MeshSyncClientBlender/msblenContextState.cpp
seandillon92: Just checked, its there on dev https://github.com/Unity-Technologies/MeshSyncDCCPlugins/blob/dev/Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp

Plugins~/Src/MeshSyncClient/Include/MeshSyncClient/msTransformManager.h
seandillon92: Should be fixed now. I used the Untabify command to fix the file

Plugins~/Src/MeshSyncClientBlender/msblenContextGeometryNodes.cpp
seandillon92: True, I simplified the branching now

Plugins~/Src/MeshSyncClientBlender/msblenGeometryNodeUtils.h
seandillon92: Fixed now, used Untabify in VS. Will apply it on all my files in another PR

seandillon92:
Moved to https://github.com/Unity-Technologies/MeshSyncDCCPlugins/pull/185 to fix issue with CLA client

PR #167, opened 2022-03-07
[Do not merge] Material baking functionality

SonnyCampbellUnity:
Support for baking Diffuse, Roughness, AO textures and syncing results to unity.

Roughness is packed into the alpha channel of the Diffuse map for use with Unity's standard shader.

[Do not merge: This PR will wait for the geometry nodes work to be complted.]


SonnyCampbellUnity:
Will not merge - feature being worked on by Christian

PR #168, opened 2022-03-08
ARTDEV-15192 fixing auto sync status when file loaded

agicpp:
Bugfix for
https://jira.unity3d.com/browse/ARTDEV-15192

The __end__ will be called after the invoke in all cases. When the file is loaded, the modal will be stopped and the __end__ is called. This will switch off the auto sync and the button will be rendered correctly (play). If the user clicks on auto sync again, the sync will continue. Tested on all blender versions which have a corresponding .py file


unity-cla-assistant:
[CLA assistant check](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/MeshSyncDCCPlugins?pullRequest=168) <br/>All committers have signed the CLA.

sindharta (approved):

PR #178, opened 2022-04-06
blender-feat: Unity->Blender communication 

schinkowski:
Adding 2 way communication between unity and blender.


unity-cla-assistant:
[CLA assistant check](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/MeshSyncDCCPlugins?pullRequest=178) <br/>All committers have signed the CLA.

Plugins~/Src/MeshSyncClient/msPropertyManager.cpp
```diff
@@ -0,0 +1,74 @@
+#include "MeshSyncClient/msPropertyManager.h"
+#include "pch.h"
+#include "MeshSync/SceneGraph/msPropertyInfo.h"
+
+using namespace std;
+
+namespace ms {
+	vector<PropertyInfoPtr> ms::PropertyManager::getAllProperties()
```
seandillon92, line 8: I think this should be named getDirtyProperties as it gets the dirty entries only

Plugins~/Src/MeshSyncClient/msPropertyManager.cpp
```diff
@@ -0,0 +1,74 @@
...
+		return m_receivedProperties;
+	}
+
+	std::vector<EntityPtr> PropertyManager::getReceivedEntities() {
+		std::unique_lock<std::mutex> lock(m_mutex);
+
+		return m_receivedEntities;
+	}
+
+	void PropertyManager::clearReceivedProperties() {
```
seandillon92: This also clears the Entities, maybe it could be renamed to reflect that.

Plugins~/Src/MeshSyncClientBlender/BlenderPyObjects/BlenderPyScene.h
```diff
@@ -6,49 +6,79 @@
...
-    {
-        for (auto *c : list_range(cc)) {
-            each_objects_impl(body, (CollectionChild*)c->collection->children.first);
-            for (auto *o : list_range((CollectionObject*)c->collection->gobject.first))
+    class BlenderPyScene {
+    public:
+        MSBLEN_BOILERPLATE2(BlenderPyScene, Scene)
+            MSBLEN_COMPATIBLE(BlenderPyID)
+
+            int fps() const;
```
seandillon92: Something is off with indentation here

Plugins~/Src/MeshSyncClientBlender/BlenderPyObjects/BlenderPyScene.h
```diff
@@ -6,49 +6,79 @@
...
+        void each_selection(const Body& body)
+        {
+            each_objects([&](Object* obj) {
+                BObject bo(obj);
+                if (bo.is_selected())
+                    body(obj);
+            });
+        }
+
+        Object* findObjectInCollection(std::string objName, Collection* col) {
```
seandillon92: we should agree on a convention but it seems that this file is using snake case for the most part. Might be good to try and keep it as consistent as we can for now

Plugins~/Src/MeshSyncClientBlender/msblenBinder.cpp
```diff
@@ -389,7 +459,35 @@ void BMesh::calc_normals_split()
...
+{
+    call<Mesh, void>(g_context, m_ptr, BMesh_update);    
+}
+
+void BMesh::clear_geometry()
+{
+    call<Mesh, void>(g_context, m_ptr, BMesh_clear_geometry);
+}
+
+void BMesh::addVertices(int count) {
```
seandillon92: The convention here is inconsistent

Plugins~/Src/MeshSyncClientBlender/msblenBinder.cpp
```diff
@@ -415,6 +513,17 @@ MLoopUV* BEditMesh::GetUV(const int index) const {
     return static_cast<MLoopUV *>(CustomData_get_layer_n(&m_ptr->bm->ldata, CD_MLOOPUV, index));
 }
 
+void BNurb::add_bezier_points(int count, Object* obj) {
+    call<Nurb, void, int>(g_context, m_ptr, BNurb_splines_bezier_add, count, obj->id);
+}
+
+void BCurve::clearSplines() {
```
seandillon92: here as well

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
```diff
@@ -59,17 +65,29 @@ void msblenContext::NodeRecord::recordAnimation(msblenContext *_this) const {
...
+static msblenContext* s_instance = nullptr;
+
 msblenContext& msblenContext::getInstance() {
-    static msblenContext s_instance;
-    return s_instance;
+    if (s_instance == nullptr) {
+        s_instance = new msblenContext();
+    }
+
+    return *s_instance;
```
seandillon92, line 64: Is this change necessary?

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
```diff
@@ -59,17 +65,29 @@ void msblenContext::NodeRecord::recordAnimation(msblenContext *_this) const {
...
+    }
+
     m_texture_manager.clear();
     m_material_manager.clear();
 
     m_entities_state->clear();
     m_instances_state->clear();
+
+    delete s_instance;
+    s_instance = nullptr;
```
seandillon92, line 79: Why do we need this?

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
```diff
@@ -644,6 +682,271 @@ ms::LightPtr msblenContext::exportLight(msblenContextState& state, msblenContext
     return ret;
 }
 
+ms::CurvePtr msblenContext::exportCurve(msblenContextState& state, msblenContextPathProvider& paths, BlenderSyncSettings& settings, const Object* src)
```
seandillon92: There is a lot of code added in this file which is already very large and does too much. Ideally we would be able to break it down in the future to make it more testable and try to reduce its responsibilities as a design improvement. As a first step, could we have another file with this code which is specific to Curves? similar to https://github.com/Unity-Technologies/MeshSyncDCCPlugins/blob/65d476985dc2ee8babe6bafa1e1372a56d9c05bd/Plugins~/Src/MeshSyncClientBlender/msblenContextGeometryNodes.cpp

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
```diff
@@ -763,6 +1066,8 @@ void msblenContext::doExtractMeshData(msblenContextState& state, BlenderSyncSett
         dst.refine_settings.flags.Set(ms::MESH_REFINE_FLAG_GEN_TANGENTS, true);
     dst.refine_settings.flags.Set(ms::MESH_REFINE_FLAG_FLIP_FACES, true);
     dst.refine_settings.flags.Set(ms::MESH_REFINE_FLAG_MAKE_DOUBLE_SIDED, settings.make_double_sided);
+
+   // dst.refine_settings.flags.Set(ms::MESH_REFINE_FLAG_FLIP_FACES, false); // test
```
seandillon92: this is dead code, should be removed

Plugins~/Src/MeshSyncClientBlender/msblenContextIntermediatePathProvider.cpp
```diff
@@ -4,7 +4,7 @@
 
 std::string append_id(std::string path, const Object* obj) {
     auto data = (ID*)obj->data;
-    path += "_" + std::to_string(data->session_uuid);
+    path += "_" + std::string(data->name); // std::to_string(data->session_uuid);
```
seandillon92: This might cause some issues with having unique paths for meshes created in Geometry Node trees. What problem is this change addressing?

Plugins~/Src/MeshSyncClientBlender/msblenUtils.cpp
```diff
@@ -3,157 +3,184 @@
...
-        ret.assign(obj->id.name + 2);
-        mu::SanitizeNodeName(ret);
+    std::string get_name(const Material* obj)
+    {
+        std::string ret;
+        if (obj) {
+            ret.assign(obj->id.name + 2);
+            mu::SanitizeNodeName(ret);
+        }
+        return ret;
```
seandillon92: Is there any change in this file apart from the style changes? If not, can it be reverted until we get a linter?

Plugins~/Src/MeshSyncClientBlender/msblenUtils.h
```diff
@@ -15,101 +15,103 @@ bool visible_in_viewport(const Object *obj);
 bool visible_in_collection(LayerCollection* layerCollection, const Object* obj);
 bool visible_in_collection(const Object* obj);
 const ModifierData* FindModifier(const Object *obj, ModifierType type);
+const ModifierData* FindModifier(const Object* obj, const std::string name);
```
seandillon92: Similarly, can we revert the file changes if its only style changes?

seandillon92 (changes requested):
Blocking until we resolve the intermediate path change

Plugins~/Src/MeshSyncClient/msPropertyManager.cpp
schinkowski: I've changed the code. We actually send all properties, there is no dirty check. These are very lightweight and don't need a dirty check.

Plugins~/Src/MeshSyncClient/msPropertyManager.cpp
schinkowski: Done

Plugins~/Src/MeshSyncClientBlender/BlenderPyObjects/BlenderPyScene.h
schinkowski: Fixed

Plugins~/Src/MeshSyncClientBlender/BlenderPyObjects/BlenderPyScene.h
schinkowski: Fixed

Plugins~/Src/MeshSyncClientBlender/msblenBinder.cpp
schinkowski: Fixed

Plugins~/Src/MeshSyncClientBlender/msblenBinder.cpp
schinkowski: Fixed

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
schinkowski: Yes because I need a reference to the instance in the Destroy method for correct clean up.

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
schinkowski: Without this the context was not destroyed at the correct time, which caused the AsyncSceneSender to not be destroyed and that kept blender alive and didn't allow it to shut down because I have the long running socket request to unity to get changed data from it. With this change, the sender's destructor is called when the context is destroyed and the sender can cancel that request in its destructor.

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
schinkowski: Can we do that as part of a refactor afterwards? We could extract all these export methods for meshes, lights, etc. to their own handler classes.

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
schinkowski: Fixed

Plugins~/Src/MeshSyncClientBlender/msblenContextIntermediatePathProvider.cpp
schinkowski: This change ensures that meshes have consistent names every sync, the session_uuid changes.
I've tested this change with every test file from testrail and also made my own test file that creates primitives inside the geo node and it works fine.

Plugins~/Src/MeshSyncClientBlender/msblenUtils.cpp
schinkowski: Fixed

Plugins~/Src/MeshSyncClientBlender/msblenUtils.h
schinkowski: Fixed

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
seandillon92: I would be in favor of continuous refactoring. Given that the C++ layer has no tests, big refactoring would have a lot of risk.

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
```diff
@@ -1320,16 +1625,44 @@ bool msblenContext::sendMaterials(bool dirty_all)
...
+void msblenContext::requestServerInitiatedMessage()
+{
+    if (m_settings.ExportSceneCache) {
+        return;
+    }
+
+    m_sender.on_server_initiated_response_received = [this](auto properties, auto entities, std::string messageFromServer) {
+        m_property_manager.updateFromServer(properties, entities);
+
+        if (messageFromServer == "sync") {
```
seandillon92: Would it be better to use a const instead of hardcoding here? Maybe have a file with server messages definitions

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
```diff
@@ -1320,16 +1625,44 @@ bool msblenContext::sendMaterials(bool dirty_all)
...
+        dirty_all = true;
+        m_server_requested_sync = false;
+    }
+
     m_settings.Validate();
     m_entity_manager.setAlwaysMarkDirty(dirty_all);
     m_material_manager.setAlwaysMarkDirty(dirty_all);
     m_texture_manager.setAlwaysMarkDirty(false); // false because too heavy
     m_instances_state->manager.setAlwaysMarkDirty(dirty_all);
+	m_property_manager.clear();
```
seandillon92: indentation issue here

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
```diff
@@ -1363,6 +1697,7 @@ bool msblenContext::sendObjects(MeshSyncClient::ObjectScope scope, bool dirty_al
 
     m_asyncTasksController.Wait();
     m_entities_state->eraseStaleObjects();
+    //m_instances_state->eraseStaleObjects();
```
seandillon92: dead code, should be removed

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
```diff
@@ -1340,6 +1673,7 @@ bool msblenContext::sendObjects(MeshSyncClient::ObjectScope scope, bool dirty_al
             return true; // nothing to send
 
         bl::BlenderPyScene scene = bl::BlenderPyScene(bl::BlenderPyContext::get().scene());
+
```
seandillon92: empty line

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
```diff
@@ -1535,6 +1870,7 @@ void msblenContext::flushPendingList(msblenContextState& state, msblenContextPat
     }
 }
 
+
```
seandillon92: empty line

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
```diff
@@ -1556,9 +1892,11 @@ void msblenContext::WaitAndKickAsyncExport()
 
     using Exporter = ms::SceneExporter;
     Exporter *exporter = m_settings.ExportSceneCache ? (Exporter*)&m_cache_writer : (Exporter*)&m_sender;
-
+
```
seandillon92, line 1658: empty line

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
```diff
@@ -1577,10 +1915,12 @@ void msblenContext::WaitAndKickAsyncExport()
         t.geometries = m_entity_manager.getDirtyGeometries();
         t.instanceInfos = m_instances_manager.getDirtyInstances();
         t.instanceMeshes = m_instances_manager.getDirtyMeshes();
+		t.propertyInfos = m_property_manager.getAllProperties();
```
seandillon92: indentation issue

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
```diff
@@ -1577,10 +1915,12 @@ void msblenContext::WaitAndKickAsyncExport()
...
         t.animations = m_animations;
 
         t.deleted_materials = m_material_manager.getDeleted();
         t.deleted_entities = m_entity_manager.getDeleted();
+
```
seandillon92: empty line

Plugins~/Src/MeshSyncClientBlender/msblenContext.h
```diff
@@ -19,25 +19,31 @@
...
 #if BLENDER_VERSION >= 300
 #include <msblenGeometryNodeUtils.h>
 #endif
 
-#include "../MeshSyncClientBlender/msblenContextState.h"
-#include <MeshSyncClient/msEntityManager.h>
 #include <msblenContextDefaultPathProvider.h>
 #include <msblenContextIntermediatePathProvider.h>
 
+
```
seandillon92: empty line

Plugins~/Src/MeshSyncClientBlender/msblenContext.h
```diff
@@ -64,6 +70,9 @@ class msblenContext {
 
     void onDepsgraphUpdatedPost(Depsgraph* graph);
 
+    void requestServerInitiatedMessage();
+
+    void callPythonMethod(const char* name);
```
seandillon92: Should this be in a Util class?

Plugins~/Src/MeshSyncClientBlender/msblenUtils.cpp
```diff
@@ -73,19 +86,24 @@ bool visible_in_viewport(const Object *obj)
...
-    for (auto collectionObject : blender::list_range((CollectionObject*)layerCollection->collection->gobject.first)) {
+    for (auto collectionObject : blender::list_range((CollectionObject*)lc->collection->gobject.first)) {
         if (collectionObject->ob == obj) {
-            if (!(layerCollection->flag & LAYER_COLLECTION_EXCLUDE)) {
+            if ((!(lc->flag & LAYER_COLLECTION_EXCLUDE)) &&
+#if BLENDER_VERSION >= 300
+                (!(lc->collection->flag & COLLECTION_HIDE_RENDER))
+#else
+                (!(lc->collection->flag & COLLECTION_RESTRICT_RENDER))
+#endif
```
seandillon92, line 98: what does this change do?

Plugins~/Src/MeshSyncClientBlender/python/3.1.0/unity_mesh_sync.py
```diff
@@ -128,6 +128,12 @@ def __del__(self):
     def execute(self, context):
         return self.invoke(context, None)
 
+    # When a file is loaded, the modal registration is reset:
+    @persistent
+    def load_handler(dummy):
+        MESHSYNC_OT_AutoSync._registered = False
+        bpy.app.handlers.load_post.remove(MESHSYNC_OT_AutoSync.load_handler)
```
seandillon92: Is this related to this PR or an independent bug fix?

Plugins~/Src/MeshSyncClient/msPropertyManager.cpp
```diff
@@ -0,0 +1,72 @@
...
+
+		return ret;
+	}
+
+	void PropertyManager::add(PropertyInfoPtr propertyInfo)
+	{
+		std::unique_lock<std::mutex> lock(m_mutex);
+
+		// Make sure the same property doesn't get added twice when objects and their parents are iterated:
+		for (size_t i = 0; i < m_records.size(); ++i) {
```
sindharta: Shouldn't we use another type of collection for this ? `std::map` for example ?

Plugins~/Src/MeshSyncClientBlender/BlenderPyObjects/BlenderPyScene.h
```diff
@@ -6,49 +6,79 @@
 #include "BlenderPyID.h" //BID
```
sindharta, line 6: there are a lot of indentation changes in this file, which should be separated in another PR if necessary.

Plugins~/Src/MeshSyncClientBlender/msblenBinder.cpp
```diff
@@ -18,6 +18,7 @@ bContext *g_context;
 extern PropertyRNA* BlenderPyID_is_updated;
 extern PropertyRNA* BlenderPyID_is_updated_data;
 extern FunctionRNA* BlenderPyID_evaluated_get;
+extern FunctionRNA* BlenderPyID_update_tag;
```
sindharta: new code for registering new Blender functions, which is relatively safe, can be put in another PR

Plugins~/Src/MeshSyncClientBlender/msblenContextIntermediatePathProvider.cpp
sindharta: @seandillon92 I think you might be more familiar in this area than I am. Please let us know what you think.

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
sindharta: Can you please put the relationship of this with `AsyncSceneSender` into a diagram so that we can understand it better ?

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
```diff
@@ -1598,7 +1938,14 @@ void msblenContext::WaitAndKickAsyncExport()
         m_entity_manager.clearDirtyFlags();
         m_animations.clear();
         m_instances_manager.clearDirtyFlags();
+
+        callPythonMethod("meshsync_post_export");
```
sindharta: Can probably be put in another PR.

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
sindharta: If it's possible to put the new codes in another file, then yes, I'd also prefer having it from the beginning.

Plugins~/Src/MeshSyncClientBlender/msblenModifiers.cpp
```diff
@@ -0,0 +1,325 @@
...
+					}
+					case IDP_FLOAT: {
+						propertyInfo->set((float*)IDP_Array(property), defaultValue->min, defaultValue->max, property->len);
+						break;
+					}
+					}
+					break;
+				}
+				case IDP_STRING: {
+					auto uiData = (bNodeSocketValueString*)property->ui_data;
```
sindharta: `uiData` is not used

Plugins~/Src/MeshSyncClientBlender/msblenModifiers.cpp
```diff
@@ -0,0 +1,325 @@
...
+	void addCustomProperties(const Object* obj, ms::PropertyManager* propertyManager) {
+		if (obj->id.properties) {
+			for (auto property : blender::list_range((IDProperty*)obj->id.properties->data.group.first)) {
+				if (property->ui_data == nullptr && property->type != IDP_STRING) {
+					continue;
+				}
+
+				auto propertyInfo = ms::PropertyInfo::create();
+				switch (property->type) {
+				case IDP_INT: {
```
sindharta: There are quite some code which are duplicate with code in `addModifierProperties()`.
Perhaps we can reduce it by moving some in a new function ?

Plugins~/Src/MeshSyncClientBlender/msblenModifiers.cpp
```diff
@@ -0,0 +1,325 @@
...
+		for (auto property : blender::list_range((IDProperty*)nodeModifier->settings.properties->data.group.first)) {
+			if (property->name == receivedProp.propertyName) {
+				setProperty(obj, property, receivedProp);
+			}
+		}
+	}
+
+	void applyCustomProperty(const Object* obj, ms::PropertyInfo& receivedProp) {
+		if (obj->id.properties) {
+			for (auto property : blender::list_range((IDProperty*)obj->id.properties->data.group.first)) {
```
sindharta: this loop is kind of duplicate with the one in `applyGeoNodeProperty()` ?
Maybe we can reduce duplicate code ?

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
```diff
@@ -644,6 +682,271 @@ ms::LightPtr msblenContext::exportLight(msblenContextState& state, msblenContext
...
+    std::shared_ptr<ms::Curve> ret = ms::Curve::create();
+
+    bl::BObject bobj(src);
+    Curve* data = nullptr;
+    data = (Curve*)src->data;
+
+    ms::Curve& dst = *ret;
+    dst.path = paths.get_path(src);
+
+    bool is_editing = false;
```
sindharta: unused variable ?

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
```diff
@@ -644,6 +682,271 @@ ms::LightPtr msblenContext::exportLight(msblenContextState& state, msblenContext
...
+            importMesh(dynamic_cast<ms::Mesh*>(entity.get()));
+            break;
+        default:
+            break;
+        }
+    }
+}
+
+// I don't think we can just mem copy here because this is also used for multidimensional arrays:
+#define copyFloatVector(DST, SRC) \
```
sindharta: what about `std::copy()` ?

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
```diff
@@ -644,6 +682,271 @@ ms::LightPtr msblenContext::exportLight(msblenContextState& state, msblenContext
...
+    std::vector<int> rev_mid_table(mid_table.size());
+    for (int i = 0; i < mid_table.size(); i++)
+    {
+        rev_mid_table[mid_table[i]] = i;
+    }
+
+    int num_indices = mesh->indices.size();
+    int num_polygons = num_indices / 3;
+
+    // The mesh for unity had split vertices, we need to undo that for blender so the faces are connected:
```
sindharta, line 621: This is a strong candidate for a function in a utility class.
I think it will also be very helpful if we can somehow have a test code with a mock test data for this.

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
schinkowski: fixed

Plugins~/Src/MeshSyncClientBlender/msblenBinder.cpp
schinkowski: Done here: https://github.com/Unity-Technologies/MeshSyncDCCPlugins/pull/208/files

Plugins~/Src/MeshSyncClientBlender/python/3.1.0/unity_mesh_sync.py
schinkowski: Unrelated to this feature, I've created a separate PR for this here: https://github.com/Unity-Technologies/MeshSyncDCCPlugins/pull/209

Plugins~/Src/MeshSyncClientBlender/msblenUtils.cpp
schinkowski: This ensures that layer visibility works the same way as object visibility, so meshsync only hides objects that are hidden in render, not viewport.

Plugins~/Src/MeshSyncClientBlender/msblenModifiers.cpp
schinkowski: Good idea, done.

Plugins~/Src/MeshSyncClient/msPropertyManager.cpp
schinkowski: Not sure how a map would help, the 'matches' methhod checks multiple fields, unless we create some kind of hash from that. I don't think it's worth implementing because the number of properties is not that big anyway.

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
schinkowski: I've refactored it and added a handler for curves. After this merge I'd like to create more handlers based on this for other entity types.

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
schinkowski: Done here: https://confluence.unity3d.com/display/ARTDEV/Bidirectional+flow

Plugins~/Src/MeshSyncClientBlender/msblenContext.h
schinkowski: fixed

Plugins~/Src/MeshSyncClientBlender/msblenContext.h
schinkowski: I've moved it to msblenbinder.

Plugins~/Src/MeshSyncClientBlender/BlenderPyObjects/BlenderPyScene.h
schinkowski: That was merged now.

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
schinkowski: There will be a mesh handler class for this after the refactor. We only need this functionality in the handler, which will be testable on its own.

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
schinkowski: Fixed!

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
schinkowski: fixed

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
schinkowski: fixed

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
schinkowski: This is on purpose for readability.

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
schinkowski: fixed

Plugins~/Src/MeshSyncClientBlender/msblenModifiers.cpp
schinkowski: I thought about it but the data structures are different. We could probably make it work with template functions and macros but then it becomes much harder to understand, no?

Plugins~/Src/MeshSyncClientBlender/msblenContextIntermediatePathProvider.cpp
```diff
@@ -2,9 +2,33 @@
...
+                found = true;
+                break;
+            }
+        }
+
+        if (!found) {
+            path += "_" + std::to_string(data->session_uuid);
+            mappedNames.insert(std::make_pair(data->name, data->session_uuid));
+        }
+    }
```
seandillon92: Why not use an unordered set tracking the session_uuids here?

Plugins~/Src/MeshSyncClientBlender/msblenModifiers.cpp
```diff
@@ -0,0 +1,327 @@
...
+				continue;
+			}
+
+			if (doesPropertyUseAttribute(property->name, nodeModifier)) {
+				continue;
+			}
+
+			auto socket = getSocketForProperty(property, group);
+
+			if (socket != nullptr) {
```
seandillon92: Can you please invert this to if (socket == nullptr) continue?

Plugins~/Src/MeshSyncClientBlender/msblenModifiers.cpp
```diff
@@ -0,0 +1,327 @@
...
+				propertyInfo->modifierName = modifier->name;
+				propertyInfo->propertyName = property->name;
+				propertyInfo->sourceType = ms::PropertyInfo::SourceType::GEO_NODES;
+				propertyManager->add(propertyInfo);
+			}
+		}
+	}
+
+	void addCustomProperties(const Object* obj, ms::PropertyManager* propertyManager) {
+		if (obj->id.properties) {
```
seandillon92: can this be inverted for an early exit if there are no properties?

Plugins~/Src/MeshSyncClientBlender/msblenModifiers.cpp
seandillon92: It would be worth trying to reduce duplication. I find it difficult to go through this part of the code and review it properly. There is a lot of nesting too.

Plugins~/Src/MeshSyncClientBlender/msblenUtils.cpp
seandillon92: This looks like a separate fix, do you think we could have it in another PR?

seandillon92 (commented):
There is a lot of nesting and duplication in the msblenModifiers file. Would be great if this can be improved

Plugins~/Src/MeshSyncClientBlender/msblenModifiers.cpp
schinkowski: Like I said, I'm not sure how much more this can be reduced without creating a mess of templates and macros. The data structures are different, we create a PropertyInfo, but one needs the defaults from the node sockets implementation classes for each type, the other uses defaults from the type specific UI classes.

Plugins~/Src/MeshSyncClientBlender/msblenModifiers.cpp
schinkowski: fixed

Plugins~/Src/MeshSyncClientBlender/msblenModifiers.cpp
schinkowski: fixed

Plugins~/Src/MeshSyncClientBlender/msblenModifiers.cpp
schinkowski: fixed

Plugins~/Src/MeshSyncClientBlender/msblenUtils.cpp
schinkowski: Done here: https://github.com/Unity-Technologies/MeshSyncDCCPlugins/pull/219/files

Plugins~/Src/MeshSyncClientBlender/msblenContextIntermediatePathProvider.cpp
schinkowski: I only want to append a session id if I already have an item with the same name. I would not know the session ids of other items with this name here without the mappedNames collection.

seandillon92: is the session id unique across all items or unique across all items with the same name?

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
schinkowski: Fixed

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
schinkowski: I can't get that to work because we also need to do implicit casting on each member of the array.

Plugins~/Src/MeshSyncClientBlender/msblenModifiers.cpp
```diff
@@ -0,0 +1,330 @@
...
+#include "BlenderPyObjects/BlenderPyContext.h"
+#include "msblenBinder.h"
+
+namespace blender {
+#if BLENDER_VERSION < 300
+void msblenModifiers::exportProperties(const Object* obj, ms::PropertyManager* propertyManager) {}
+void msblenModifiers::importProperties(std::vector<ms::PropertyInfo> props) {}
+#else
+
+// Copied from blender source that we cannot include:
```
mloccy, line 15: Can we please be more specific as to why we cannot include this, instead of leaving it up to the reader to potentially jump to the wrong conclusion like I did.

sindharta: ~~Because we are not compiling all of Blender source files.
We are only using Blender's include files, and some of its implementation files.
And for this to work, we hand picked the implementation code that we need and put them here.
A bit messy, yes. 
I tried to include all Blender source files before but it wasn't as easy as I thought.~~


~~FYI: @schinkowski~~
~~This change is actually taken from another PR, and has been merged to dev~~
~~That is why I think we should do rebase before doing a code review, or otherwise it would show up like this occasionally.~~

Plugins~/Src/MeshSyncClientBlender/msblenContextIntermediatePathProvider.cpp
schinkowski: it's unique across all items

seandillon92: There is a potential bug in this code. Assume the following objects ask for the intermediate path:
name = "cube", id = 123
name = "cube", id = 123
name = "cube", id = 456
name = "cube", id = 456

The outputs will be 
"cube"
"cube"
"cube_456"
"cube"

and it should be 
"cube"
"cube"
"cube_456"
"cube_456"

seandillon92: Another concern is the scope and ownership of the cache. As it stands, if we have multiple geonode trees, we will get cache hits only for the tree that was executed first. My suggestion is to extract this into a cache object that is owned by the blender context class (or, in the future ,by the appropriate exported/handler class for geonodes). The owner can use the cache to decide which provider to use. This will keep this provider simple with a single responsibility (the caching part is not reflected anywhere in the name or comments).

seandillon92 (changes requested):
Requesting changes for the potential bug. The comment on cache ownership is not a blocker

seandillon92 (approved):

Plugins~/Src/MeshSyncClientBlender/msblenContextIntermediatePathProvider.cpp
```diff
@@ -2,9 +2,24 @@
 #include "msblenUtils.h"
 #include <BLI_listbase.h>
 
-std::string append_id(std::string path, const Object* obj) {
+std::string msblenContextIntermediatePathProvider::append_id(std::string path, const Object* obj) {
     auto data = (ID*)obj->data;
-    path += "_" + std::to_string(data->session_uuid);
+    
+    path += "_" + std::string(data->name);
```
seandillon92, line 8: is this necessary?

Plugins~/Src/MeshSyncClientBlender/msblenCurveHandler.cpp
```diff
@@ -0,0 +1,119 @@
...
+		else
+			task();
+	}
+
+	return ret;
+}
+
+void msblenCurveHandler::doExtractCurveData(msblenContextState& state, BlenderSyncSettings& settings, ms::Curve& dst, const Object* obj, Curve* data, mu::float4x4 world)
+{
+	if (settings.sync_curves) {
```
seandillon92, line 42: can you change this to check for the negation and return if true? will help with reducing nesting

Plugins~/Src/MeshSyncClientBlender/msblenCurveHandler.cpp
```diff
@@ -0,0 +1,119 @@
...
+			curveSpline->closed = (nurb->flagu & CU_NURB_CYCLIC) != 0;
+
+			curveSpline->cos.resize_discard(nurb->pntsu);
+			curveSpline->handles_left.resize_discard(nurb->pntsu);
+			curveSpline->handles_right.resize_discard(nurb->pntsu);
+
+			for (int i = 0; i < nurb->pntsu; i++) {
+				BezTriple* bezt = &nurb->bezt[i];
+
+				if (bezt) {
```
seandillon92: this can be changed to if(!bezt)continue to reduce nesting

Plugins~/Src/MeshSyncClientBlender/msblenModifiers.cpp
sindharta: I apologize. It seems that I had a different file in mind when commenting on this yesterday.

schinkowski: The blender source we include has been modified to avoid pulling in lots of dependencies. I just need a way to cast the property data the same way blender does it.

Plugins~/Src/MeshSyncClientBlender/msblenUtils.cpp
```diff
@@ -64,6 +65,18 @@ std::string get_path(const Object *arm, const Bone *obj)
...
+Object* get_object_from_path(std::string path) {
+    auto lastIndexOfDivider = path.find_last_of('/');
+    if (lastIndexOfDivider < 0) {
+        return nullptr;
+    }
+
+    auto objName = path.substr(lastIndexOfDivider + 1);
+
+    bl::BlenderPyScene scene = bl::BlenderPyScene(bl::BlenderPyContext::get().scene());
+    return scene.get_object_by_name(objName);
```
sindharta, line 77: Can't Blender have two objects with the same names, but different paths ?

Plugins~/Src/MeshSyncClientBlender/msblenUtils.cpp
sindharta: Please rebase/apply that merged fix to this PR.

Plugins~/Src/MeshSyncClientBlender/MeshSyncClientBlender.cpp
```diff
@@ -170,7 +170,7 @@ PYBIND11_MODULE(MeshSyncClientBlender, m)
             BindMethod(Destroy, [](self_t& self) { self->Destroy(); })
             BindMethod(setup, [](self_t& self, py::object ctx) { bl::setup(ctx); })
             BindMethod(clear, [](self_t& self) { self->clear(); })
-            BindMethod(exportUpdatedObjects, [](self_t& self) { self->sendObjects(MeshSyncClient::ObjectScope::Updated, false); })
+            BindMethod(exportUpdatedObjects, [](self_t& self) { self->sendObjects(MeshSyncClient::ObjectScope::Updated, false); self->requestServerInitiatedMessage(); })
```
sindharta: We should probably separate this into a separate method, or change the name of the bound method.

Also, `requestServerInitiatedMessage` is old. Have you reflected your changes on the MeshSync side ?

Plugins~/Src/MeshSyncClientBlender/msblenContextIntermediatePathProvider.cpp
schinkowski: Yes because the uuid changes every execution

Plugins~/Src/MeshSyncClientBlender/msblenCurveHandler.cpp
schinkowski: done

Plugins~/Src/MeshSyncClientBlender/msblenCurveHandler.cpp
schinkowski: done

Plugins~/Src/MeshSyncClientBlender/msblenUtils.cpp
schinkowski: No, the path just defines the parenting, the object names are still unique.

Plugins~/Src/MeshSyncClientBlender/MeshSyncClientBlender.cpp
schinkowski: Done but now we need this to be merged: https://github.com/unity3d-jp/MeshSync/pull/722

Plugins~/Src/MeshSyncClientBlender/msblenContextIntermediatePathProvider.cpp
seandillon92: the name will be the same right? why are we appending it?

schinkowski: Initially the path has the object name on it, this appends the mesh name, so it would be something like <object_mesh>.
This ensures that geometry node with multiple meshes with different names work without a problem.

seandillon92: Aha ok, thanks that's more clear now

seandillon92 (approved):

PR #207, opened 2022-06-03
Prep for bidirectional mesh support.

schinkowski:
Fix for python scripts so the operators can be called from command line args.
Reducing update time step for faster sync.
Preparing meshsync python events.


Plugins~/Src/MeshSyncClientBlender/python/3.0.0/unity_mesh_sync.py
```diff
@@ -0,0 +1,327 @@
+bl_info = {
```
schinkowski: @sindharta Do we support blender 3.0.0? If not I can delete this file. It is not in the cmakelist so maybe it's not supported or the same as blender 3.0.1?

Plugins~/Src/MeshSyncClientBlender/python/3.1.0/unity_mesh_sync.py
```diff
@@ -14,6 +14,17 @@
...
 
+# Events that get called during the meshsync export stages, append your functions to these lists:
+# Called every frame when checking if something needs exporting:
+mesh_sync_on_prepare = []
+
+# Called before exporting
+mesh_sync_on_pre_export = []
+
+# Called after export is finished
+mesh_sync_on_post_export = []
```
seandillon92, line 25: Are these invoked anywhere?

Plugins~/Src/MeshSyncClientBlender/python/3.0.1/unity_mesh_sync.py
```diff
@@ -113,15 +114,26 @@ class MESHSYNC_OT_AutoSync(bpy.types.Operator):
...
+
     def invoke(self, context, event):
         scene = bpy.context.scene
         if not MESHSYNC_OT_AutoSync._timer:
             scene.meshsync_auto_sync = True
             if not scene.meshsync_auto_sync:
                 # server not available
                 return {'FINISHED'}
-            MESHSYNC_OT_AutoSync._timer = context.window_manager.event_timer_add(1.0 / 3.0, window=context.window)
+            update_step = 0.0001 # 1.0/3.0
```
seandillon92, line 127: Auto sync executes a diff check on every scene object. For geometry nodes exporting instances, this is probably fine but I wonder how it will perform for projects with large scenes or a lot of geometry

Plugins~/Src/MeshSyncClientBlender/python/3.0.1/unity_mesh_sync.py
```diff
@@ -139,6 +151,8 @@ def update(self):
         msb_apply_scene_settings()
         msb_context.setup(bpy.context);
         msb_context.exportUpdatedObjects()
+    
+
```
seandillon92, line 155: These empty lines can be removed

Plugins~/Src/MeshSyncClientBlender/python/3.0.1/unity_mesh_sync.py
```diff
@@ -25,6 +25,7 @@ class MESHSYNC_PT_Main(MESHSYNC_PT, bpy.types.Panel):
 
     def draw(self, context):
         pass
+
```
seandillon92, line 28: This one can be removed

Plugins~/Src/MeshSyncClientBlender/python/3.0.0/unity_mesh_sync.py
sindharta: For the same major-minor combined version of Blender, we "officially" support only a single patch version of that combination.

In theory, what we have in 3.0.1 should work for 3.0.0 though, although we make no guarantees.

Plugins~/Src/MeshSyncClientBlender/python/3.1.0/unity_mesh_sync.py
schinkowski: Not yet, the code to invoke them will be in a later PR.

Plugins~/Src/MeshSyncClientBlender/python/3.0.0/unity_mesh_sync.py
schinkowski: Ok, I've removed this file.

Plugins~/Src/MeshSyncClientBlender/python/3.0.1/unity_mesh_sync.py
schinkowski: It runs in its own thread. I've tested with heavy scenes and didn't see any performance impact until the scene changed, so the check for changes seems to be lightweight enough.

sindharta (approved):

sindharta:
@seandillon92 Did you have all your comments addressed for this PR ?

seandillon92 (approved):

seandillon92:
> @seandillon92 Did you have all your comments addressed for this PR ?

Yes, should be good for merging

PR #209, opened 2022-06-23
blender-fix: make AutoSync work again after loading a new file

schinkowski:
Ensure python callback is only registered once to avoid getting too many callbacks.

### DIFF
--- a/Plugins~/Src/MeshSyncClientBlender/python/3.1.0/unity_mesh_sync.py
+++ b/Plugins~/Src/MeshSyncClientBlender/python/3.1.0/unity_mesh_sync.py
@@ -132,7 +132,6 @@ class MESHSYNC_OT_AutoSync(bpy.types.Operator):
     @persistent
     def load_handler(dummy):
         MESHSYNC_OT_AutoSync._registered = False
-        bpy.app.handlers.load_post.remove(MESHSYNC_OT_AutoSync.load_handler)
 
     def invoke(self, context, event):
         scene = bpy.context.scene
@@ -149,7 +148,6 @@ class MESHSYNC_OT_AutoSync(bpy.types.Operator):
             if not MESHSYNC_OT_AutoSync._registered:
                 context.window_manager.modal_handler_add(self)
                 MESHSYNC_OT_AutoSync._registered = True
-                bpy.app.handlers.load_post.append(MESHSYNC_OT_AutoSync.load_handler)
 
             if bpy.app.background:
                 import time
@@ -323,6 +321,7 @@ classes = (
 
 def register():
     msb_initialize_properties()
+    bpy.app.handlers.load_post.append(MESHSYNC_OT_AutoSync.load_handler)
     for c in classes:
         bpy.utils.register_class(c)
 
--- a/Plugins~/Src/MeshSyncClientBlender/python/3.1.0/unity_mesh_sync.py
+++ b/Plugins~/Src/MeshSyncClientBlender/python/3.1.0/unity_mesh_sync.py
@@ -128,6 +128,12 @@ class MESHSYNC_OT_AutoSync(bpy.types.Operator):
     def execute(self, context):
         return self.invoke(context, None)
 
+    # When a file is loaded, the modal registration is reset:
+    @persistent
+    def load_handler(dummy):
+        MESHSYNC_OT_AutoSync._registered = False
+        bpy.app.handlers.load_post.remove(MESHSYNC_OT_AutoSync.load_handler)
+
     def invoke(self, context, event):
         scene = bpy.context.scene
         if not MESHSYNC_OT_AutoSync._timer:
@@ -143,6 +149,7 @@ class MESHSYNC_OT_AutoSync(bpy.types.Operator):
             if not MESHSYNC_OT_AutoSync._registered:
                 context.window_manager.modal_handler_add(self)
                 MESHSYNC_OT_AutoSync._registered = True
+                bpy.app.handlers.load_post.append(MESHSYNC_OT_AutoSync.load_handler)
 
             if bpy.app.background:
                 import time
--- a/Plugins~/Src/MeshSyncClientBlender/python/2.83.18/unity_mesh_sync.py
+++ b/Plugins~/Src/MeshSyncClientBlender/python/2.83.18/unity_mesh_sync.py
@@ -109,18 +109,18 @@ class MESHSYNC_OT_AutoSync(bpy.types.Operator):
     bl_idname = "meshsync.auto_sync"
     bl_label = "Auto Sync"
     _timer = None
-    
+    _registered = False
+
     def __del__(self):
         MESHSYNC_OT_AutoSync._timer = None
 
     def execute(self, context):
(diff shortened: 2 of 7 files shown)


Plugins~/Src/MeshSyncClientBlender/python/3.1.0/unity_mesh_sync.py
```diff
@@ -143,6 +149,7 @@ def invoke(self, context, event):
             if not MESHSYNC_OT_AutoSync._registered:
                 context.window_manager.modal_handler_add(self)
                 MESHSYNC_OT_AutoSync._registered = True
+                bpy.app.handlers.load_post.append(MESHSYNC_OT_AutoSync.load_handler)
```
seandillon92: _registered is reset on file load? the root cause is not clear to me, the comments above seem to acknowledge the issue.

seandillon92 (commented):
Should this change be backported to previous versions of Blender?

Plugins~/Src/MeshSyncClientBlender/python/3.1.0/unity_mesh_sync.py
schinkowski: Blender clears the modal handler callback when a file is loaded, with this change we know when a file is loaded and the handler is cleared and can register it again. The API doesn't give us a way to remove handlers or check whether or not they are registered so this is the only way we can deal with it.

seandillon92: That makes sense , thank you!

seandillon92 (approved):

Plugins~/Src/MeshSyncClientBlender/python/3.1.0/unity_mesh_sync.py
sindharta: So, I tried the sample modal operator in this [link](https://docs.blender.org/api/current/bpy.types.Operator.html) (under "Modal Execution"), and it seems that the modal handler is unregistered when we return `FINISHED` or `CANCELLED`.
Can you please confirm this ? And if so, maybe we just need to return FINISHED when opening a new blend file, rather than using additional _registered check ?

Plugins~/Src/MeshSyncClientBlender/python/3.1.0/unity_mesh_sync.py
```diff
@@ -128,6 +128,12 @@ def __del__(self):
     def execute(self, context):
         return self.invoke(context, None)
 
+    # When a file is loaded, the modal registration is reset:
```
sindharta, line 131: We should also apply these changes to other `unity_mesh_sync.py` files

Plugins~/Src/MeshSyncClientBlender/python/3.1.0/unity_mesh_sync.py
schinkowski: This operator always returns {'PASS_THROUGH'} from its 'modal' method, so it's never unregistered.

Plugins~/Src/MeshSyncClientBlender/python/3.1.0/unity_mesh_sync.py
schinkowski: Done

Plugins~/Src/MeshSyncClientBlender/python/3.1.0/unity_mesh_sync.py
sindharta: Hm, we can make it return FINISHED based on a certain variable which is set by `MESHSYNC_OT_AutoSync.load_handler`, but I guess this is just another way of doing things.

BTW, since it seems that we always need to set `_registered` to false  in `load_post`, wouldn't it be simpler to do:
```
bpy.app.handlers.load_post.append(MESHSYNC_OT_AutoSync.load_handler)
```

once when loading MeshSync plugin and never remove it ?
Putting in in `register`/`unregister` is also another alternative.

schinkowski: If we make it return FINISHED, the auto-sync would stop working. It needs to return PASS_THROUGH so that it gets called until autosync is disabled. 
We only want to register the handler when auto-sync is on, if I add a handler to load_post, that would also trigger when a file is reloaded but I just want to set the flag to know whether it's registered or not, loading a file also resets modal operators, so we would not know whether or not it shouuld be running and users probably don't want to keep this running when a new file is loaded anyway.

sindharta: > We only want to register the handler when auto-sync is on

Why ?
As far as I understand, we have the following cases:

1. Initially, `_registered` is false
2. If auto-sync is on, then we need to register `load_handler` to set `_registered` to false when opening a new blend file, because as you said, loading a file resets modal operators

So the question is, will there be cases where we need to keep `_registered` to true when opening a new blend file ? And if not, won't it be simpler to just register `load_handler` once during the plugin load ?

schinkowski: Ah, so you mean remove the _registered flag altogether and just call context.window_manager.modal_handler_add(self) when a file is loaded and a new operator is created?
I've tried that but we have no reference to the operator in the load callback, it's a class method. Also, when the operator is created, we cannot register in its init method because we can't access the context. Operators are supposed to do stuff in their execute method.

sindharta: No, I meant that we could keep using `load_handler` but only set it up once during the entire Blender session.

```
bpy.app.handlers.load_post.append(MESHSYNC_OT_AutoSync.load_handler)
```

I guess there are a couple of places which would be good candidates to put this code:
1. [near the end of the py script](https://github.com/Unity-Technologies/MeshSyncDCCPlugins/blob/31c5f6599885f6a6fa07c9b740f1306eb385f8cc/Plugins~/Src/MeshSyncClientBlender/python/3.1.0/unity_mesh_sync.py#L353)
2. in [register() ](https://github.com/Unity-Technologies/MeshSyncDCCPlugins/blob/31c5f6599885f6a6fa07c9b740f1306eb385f8cc/Plugins~/Src/MeshSyncClientBlender/python/3.1.0/unity_mesh_sync.py#L317). If we want to ensure the cleanup, we can also remove `load_handler` in [unregister()](https://github.com/Unity-Technologies/MeshSyncDCCPlugins/blob/31c5f6599885f6a6fa07c9b740f1306eb385f8cc/Plugins~/Src/MeshSyncClientBlender/python/3.1.0/unity_mesh_sync.py#L322) as well, although I don't think this is necessary.

And a friendly reminder to please replicate the updates to other py files if you make these changes.

schinkowski: Ok, I've changed it to do that instead.

PR #217, opened 2022-07-04
blender-feat: support for Blender 3.2.0

sindharta:


Plugins~/Src/MeshSyncClientBlender/BlenderSrc/material.cpp
```diff
@@ -10,6 +10,7 @@
 #undef rad2 //required before including "DNA_meta_types.h" 
 #include "DNA_meta_types.h" //MetaBall
 
+#if BLENDER_VERSION < 302
```
sindharta, line 13: As written as the head of the file, this file contains copies of  some code taken from "source\blender\blenkernel\intern\material.c".
And Blender 3.2.0 contains some differences from earlier versions.

I don't want to diverge too much from the original code at this point, so the code style might be different a lot from ours

Plugins~/Src/MeshSyncClientBlender/pch.h
```diff
@@ -48,7 +48,9 @@ namespace py = pybind11;
 #include "DNA_camera_types.h"
 #include "DNA_collection_types.h"
 #include "DNA_gpencil_types.h" //bGPdata
+#if BLENDER_VERSION < 302
 #include "DNA_hair_types.h" //Hair
```
sindharta, line 52: Hair has been removed from Blender 3.2

Plugins~/Src/MeshSyncClientBlender/CMakeLists.txt
```diff
@@ -20,6 +20,7 @@ set(BLENDER_INCLUDE_DIRS
     "${BLENDER${BLENDER_VERSION}_INCLUDE_DIR}/makesdna"
     "${BLENDER${BLENDER_VERSION}_INCLUDE_DIR}/makesrna"
     "${BLENDER${BLENDER_VERSION}_INCLUDE_DIR}/python"
+    "${BLENDER${BLENDER_VERSION}_INCLUDE_DIR}/../../intern/guardedalloc" #for Blender 3.2+
```
seandillon92: Which class depends on this?

seandillon92 (approved):

Plugins~/Src/MeshSyncClientBlender/CMakeLists.txt
sindharta: `source\blender\blenlib\BLI_memory_utils.hh`

sindharta: which is required by `BKE_customdata.h` in Blender 3.2

PR #218, opened 2022-07-04
blender-change: import directly from unity_mesh_sync.py and remove __init__.py

schinkowski:
Changing blender addon to put all its files into its own folder in the blender addon directory.


sindharta:
@seandillon92 Can you please give your review ?

sindharta (approved):

seandillon92 (changes requested):
Looks ok, but you missed the 3.2 script. Requesting changes as it would block users from installing on 3.2

schinkowski:
> Looks ok, but you missed the 3.2 script. Requesting changes as it would block users from installing on 3.2

Updated the file for 3.2 now, we didn't have 3.2 when I first did this.

seandillon92 (approved):

PR #223, opened 2022-07-07
blender-feat: also send meshes that are particle instances

schinkowski:


Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
```diff
@@ -647,11 +647,6 @@ ms::LightPtr msblenContext::exportLight(msblenContextState& state, msblenContext
 
 ms::MeshPtr msblenContext::exportMesh(msblenContextState& state, msblenContextPathProvider& paths, BlenderSyncSettings& settings, const Object *src)
 {
-    // ignore particles
-    if (//FindModifier(src, eModifierType_ParticleSystem) ||
-        FindModifier(src, eModifierType_ParticleInstance) )
-        return nullptr;
-
```
seandillon92, line 654: Wonder why this existed in the first place.. Maybe we could have a toggle to sync particle systems like the Bake Modifiers option, to avoid disrupting existing workflows

sindharta: For reference: it seems that this check was added because it required further implementation
https://github.com/unity3d-jp/MeshSync/commit/96c10275bc4f4325c1483f1a7bf2cfffc2e7d132

So, are you saying that we can sync particle animations now by just removing these lines ?

schinkowski: I don't know about particle animations. Do we have a test or a blend file for that?

sindharta: I don't, at least. I thought you had one.
How did you test your change then ?

schinkowski: I got a file from Erin which required this to work. I also made simplified test file. The particle instances are baked into a single mesh when the 'bake modifiers' option is enabled. Particle animations work and are baked into a mesh every frame. If this causes issues we could make it optional to send particle instances.

sindharta (approved):

seandillon92 (approved):

PR #224, opened 2022-07-08
blender-fix: ceil floats to avoid hashing issues

schinkowski:
Round normals to 2 decimal places to avoid issues with meshes being sent every frame


sindharta:
We would need to get https://github.com/Unity-Technologies/MeshSyncDCCPlugins/pull/178 merged first before reviewing this.

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
```diff
@@ -905,8 +905,9 @@ void msblenContext::doExtractNonEditMeshData(msblenContextState& state, BlenderS
         blender::barray_range<mu::tvec3<float>> normals = bmesh.normals();
         if (!normals.empty()) {
             dst.normals.resize_discard(num_indices);
+            const float roundingFactor = 100.0f;
             for (size_t ii = 0; ii < num_indices; ++ii)
-                dst.normals[ii] = normals[ii];
+                dst.normals[ii] = ceil(normals[ii] * roundingFactor) / roundingFactor;
```
SonnyCampbellUnity: Can you create a utility function for this in the msblenUtils class or something like that? Seems like this logic would be commonly re-used

schinkowski: Done here https://github.com/unity3d-jp/MeshSync/pull/710

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
```diff
@@ -901,7 +901,7 @@ void msblenContext::doExtractNonEditMeshData(msblenContextState& state, BlenderS
         if (!normals.empty()) {
             dst.normals.resize_discard(num_indices);
             for (size_t ii = 0; ii < num_indices; ++ii)
-                dst.normals[ii] = normals[ii];
+                dst.normals[ii] = ms::roundToDecimals(normals[ii]);
```
schinkowski: Also needs this to work: https://github.com/unity3d-jp/MeshSync/pull/710/files

SonnyCampbellUnity (approved):

seandillon92 (approved):

PR #239, opened 2022-07-28
blender-feat: Giving server a way to request a user script callback in blender.

schinkowski:
Dependency here: https://github.com/unity3d-jp/MeshSync/pull/737


Plugins~/Src/MeshSyncClientBlender/python/3.3.0/unity_mesh_sync.py
```diff
@@ -25,6 +25,8 @@
 # Called after export is finished
 mesh_sync_on_post_export = []
 
+# Called when the server requests a python callback
+mesh_sync_server_requested_callback = []
```
seandillon92: how is this array populated?

Plugins~/Src/MeshSyncClientBlender/python/3.3.0/unity_mesh_sync.py
```diff
@@ -380,6 +382,10 @@ def meshsync_post_export():
     for f in mesh_sync_on_post_export:
         f()
 
+def meshsync_server_requested_callback():
+    for f in mesh_sync_server_requested_callback:
+        f()
```
seandillon92: might be easier to move this code to another file, perhaps unity_meshsync_common or a new module, so we don't have to edit multiple files for one code change

Plugins~/Src/MeshSyncClientBlender/python/3.1.0/unity_mesh_sync.py
```diff
@@ -25,6 +25,8 @@
 # Called after export is finished
```
seandillon92: we are missing all the versions before 3.1, is this intended?

schinkowski: Yes, this is mostly for Bay's team, who are using later blender versions and I doubt anyone else will use this in the near future while the older versions are supported.

Plugins~/Src/MeshSyncClientBlender/python/3.3.0/unity_mesh_sync.py
schinkowski: This can be populated by other python code.

Plugins~/Src/MeshSyncClientBlender/python/3.3.0/unity_mesh_sync.py
schinkowski: Done now

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
```diff
@@ -1461,6 +1464,19 @@ bool msblenContext::sendObjects(MeshSyncClient::ObjectScope scope, bool dirty_al
 
     m_property_manager.clearReceivedData();
 
+    // If a python callback was requested, that may change the scene so run it and don't export until next update:
+    if (m_server_requested_python_callback) {
```
seandillon92, line 1472: Could the python_callback be renamed to client_callback? The comment applies for all the naming in this PR. I think the purpose if this functionality it to allow the server to ask the client for invoking custom callbacks, if it's in python or not sounds irrelevant

schinkowski: Renamed it now

seandillon92 (dismissed):

sindharta (dismissed):

sindharta (dismissed):

seandillon92 (dismissed):

seandillon92 (approved):

PR #243, opened 2022-07-29
3dsmax-fix: export vertex alpha 

sindharta:
[skip ci]


sindharta:
https://github.com/unity3d-jp/MeshSync/issues/716

PR #247, opened 2022-08-10
blender-fix: fix instance scaling

schinkowski:


sindharta (approved):

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
```diff
@@ -1695,8 +1695,9 @@ void msblenContext::WaitAndKickAsyncExport()
 
                     for (size_t i = 0; i < obj->transforms.size(); ++i)
                     {
-                        // We can divide the w component instead of applying the multiplier on xyz:
-                        obj->transforms[i][3][3] /= scale_factor;
+                        obj->transforms[i][3][0] *= scale_factor;
```
sindharta, line 1698: Can you remind me why we need to do this instead of dividing the `w` component ?

seandillon92 (approved):

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
schinkowski: The w component worked but causes an error log in Unity 2021, I think it expects that component to be 1.

sindharta (approved):

PR #250, opened 2022-08-23
blender-fix: export geometry node instances when frame changes

schinkowski:


Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
```diff
@@ -1448,10 +1449,19 @@ bool msblenContext::sendObjects(MeshSyncClient::ObjectScope scope, bool dirty_al
     }
 
 #if BLENDER_VERSION >= 300
-    if (m_geometryNodeUtils.getInstancesDirty() || dirty_all) {
+    // The dependency graph update event does not fire when the scene changes while the timeline is playing.
+    // To ensure geometry nodes get exported correctly while playing, check the frame number:
+    static int lastExportedFrame = scene.GetCurrentFrame();
+    int currentFrame = scene.GetCurrentFrame();
```
seandillon92: Just a concern for the future with this paradigm:
what is the lifetime of lastExportedFrame?
If we start writing unit tests, would we have to reset the variable before each tests? how would this work?

schinkowski: Lifetime is for the duration of the program. A unit test would need to take this into account and treat it like a meshsync session where frame updates trigger an export. To reset the frame variable, it could set the scene frame and call this.

seandillon92: > To reset the frame variable, it could set the scene frame and call this

Would also require to force it somehow to get through the if statement bellow. 

Would the var need to be reset as well when users open a new project? 
Might be better we have a small class that deals with this logic, holds an internal state with the counters and can be reset on an event on the blender context?

schinkowski: The first sync would be a full sync and sync instances anyway so it's not a problem if the first value is not the current frame.

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
```diff
@@ -1448,10 +1449,19 @@ bool msblenContext::sendObjects(MeshSyncClient::ObjectScope scope, bool dirty_al
     }
 
 #if BLENDER_VERSION >= 300
-    if (m_geometryNodeUtils.getInstancesDirty() || dirty_all) {
+    // The dependency graph update event does not fire when the scene changes while the timeline is playing.
+    // To ensure geometry nodes get exported correctly while playing, check the frame number:
+    static int lastExportedFrame = scene.GetCurrentFrame();
```
sindharta: Confirming.
Is this assignment of `scene.GetCurrentFrame` an optimization to skip export when doing the first sync ?
I am wondering why we are not simply assigning -1 or something else into `lastExportedFrame`.

schinkowski: It could probably be anything because the first sync will be a full sync anyway. I used GetCurrentFrame() to initialize it to the same value as the currentFrame variable.

sindharta: If that is the case, can't we just set it with -1 to: 
1. clearly indicate that the initial sync is supposed to be a full sync 
2. avoid unnecessary function call

schinkowski: Ok done

sindharta:
@seandillon92 Can you please give your review ?

seandillon92 (approved):

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
seandillon92: > A unit test would need to take this into account and treat it like a meshsync session where frame updates trigger an export. To reset the frame variable, it could set the scene frame and call this.

This would block us from running tests in parallel. I don't mind the particular example here, and so I approved this PR, but I am more concerned about repeating this pattern across the codebase.

PR #251, opened 2022-08-24
change: change plugins folder to Plugins to reveal it in the Unity Editor

seandillon92:
Changed name of folder that containts the Plugin zip files from "Plugins~" to "Plugins". The "~" character at the end of a folder name, excludes the folder from importation in the Editor.


seandillon92:
@sindharta-tanuwijaya I added a symlink from Plugins~ to Plugins. Tested with the Install Plugin button and it seems to be working well. Do you think we should keep the "Plugins~" folder in the paths mentioned in docs or replace it with "Plugins"?

sindharta:
For this package, we can just replace it with `Plugins`

seandillon92:
@schinkowski @SonnyCampbellUnity @mloccy another PR that a few others depend on. Would be good to review this one first

SonnyCampbellUnity (approved):

sindharta (approved):

PR #255, opened 2022-08-30
blender-fix: make all object types (incl. lights) instantiable for geometry nodes 

seandillon92:
Support geometry node graphs that use lights instead of meshes as input.


sindharta:
Waiting for a fix on the MeshSync side

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
```diff
@@ -740,9 +740,12 @@ ms::MeshPtr msblenContext::exportMesh(msblenContextState& state, msblenContextPa
                 Depsgraph* depsgraph = bl::BlenderPyContext::get().evaluated_depsgraph_get();
                 bobj = (Object*)bl::BlenderPyID(bobj).evaluated_get(depsgraph);
             }
-            if (Mesh *tmp = bobj.to_mesh()) {
+            if (Mesh* tmp = bobj.to_mesh()) {
                 data = tmp;
-                m_meshes_to_clear.push_back(src);
+                // Only clear meshes that aren't created from others:
+                if (src->id.orig_id == NULL) {
```
sindharta, line 764: Is this related to this particular fix/feat ?

Plugins~/Src/MeshSyncClientBlender/msblenContextGeometryNodes.cpp
```diff
@@ -72,19 +72,19 @@ void msblenContext::exportInstances() {
 }
 void msblenContext::exportInstancesFromFile(Object* instancedObject, Object* parent, SharedVector<mu::float4x4> mat, mu::float4x4& inverse)
 {
-    mu::parallel_for(0, mat.size(), 10, [this, &mat, &inverse](int i)
+    mu::parallel_for(0, mat.size(), 10, [this, &instancedObject, &mat, &inverse](int i)
```
sindharta: `inverse` is not used anymore. We can probably remove it ?

Plugins~/Src/MeshSyncClientBlender/msblenGeometryNodesUtils.cpp
```diff
@@ -33,21 +35,34 @@ namespace blender {
...
+        if (msblenUtils::is_camera(obj) || msblenUtils::is_light(obj)) {
+             msblenEntityHandler::applyCorrectionIfNeeded(obj, result);
+        }
+
+        result = m_blender_to_unity_world *
+            result *
             m_blender_to_unity_local;
-    }
+        
+        if (msblenUtils::is_camera(obj) || msblenUtils::is_light(obj)) {
```
sindharta, line 60: It seems that wee are doing the correction twice:
One before applying the world matrix, and another one after.
Does it have to be like this, and is it possible to combine them ?

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
schinkowski: I think it's related, it didn't happen until we tried to instance lights.

Plugins~/Src/MeshSyncClientBlender/msblenContextGeometryNodes.cpp
schinkowski: It's used again now.

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
seandillon92: I am not sure I follow the comment. What is others?

Plugins~/Src/MeshSyncClientBlender/msblenGeometryNodesUtils.cpp
```diff
@@ -33,21 +35,34 @@ namespace blender {
...
     /// <param name="blenderMatrix"></param>
     /// <returns></returns>
-    float4x4 GeometryNodesUtils::blenderToUnityWorldMatrix(const float4x4& blenderMatrix) {            
+    float4x4 GeometryNodesUtils::blenderToUnityWorldMatrix(const Object* obj, const float4x4& blenderMatrix) const
+    {        
+        float4x4 result = blenderMatrix;
+        
+        if (msblenUtils::is_camera(obj) || msblenUtils::is_light(obj)) {
+             msblenEntityHandler::applyCorrectionIfNeeded(obj, result);
+        }
```
seandillon92: applyCorrectionIfNeeded will check if the object is camera or light. The if statement should be removed

Plugins~/Src/MeshSyncClientBlender/msblenGeometryNodeUtils.h
```diff
@@ -40,10 +40,10 @@ class GeometryNodesUtils
     void each_instanced_object(std::function<void(Object*, Object*, SharedVector<mu::float4x4>, bool)> handler);
 
     /// <summary>
-    /// /// Converts the world matrix from blender to Unity coordinate system
-    /// /// </summary>
-    mu::float4x4 blenderToUnityWorldMatrix(const mu::float4x4& blenderMatrix);
-
+    /// Converts the world matrix from blender to Unity coordinate system
+    /// </summary>
+    mu::float4x4 blenderToUnityWorldMatrix(const Object* obj, const mu::float4x4& blenderMatrix) const;
```
seandillon92, line 45: I think passing an object type instead of the object might be better here. Should remove the requirement to keep a reference to the object.

Plugins~/Src/MeshSyncClientBlender/msblenContextGeometryNodes.cpp
```diff
@@ -45,48 +45,39 @@ void msblenContext::exportInstances() {
...
-    mu::parallel_for(0, mat.size(), 10, [this, &mat, &inverse](int i)
-        {
-            mat[i] = m_geometryNodeUtils.blenderToUnityWorldMatrix(mat[i] * inverse);
-        });
 
-    exportInstanceInfo(*m_instances_state, m_default_paths, instancedObject, parent, std::move(mat));
-}
-
-void msblenContext::exportInstancesFromTree(Object* instancedObject, Object* parent, SharedVector<mu::float4x4> mat)
+void msblenContext::exportInstancesWithPathProvider(Object* instancedObject, Object* parent, SharedVector<mu::float4x4> mat, mu::float4x4& inverse, msblenContextPathProvider& pathProvider)
```
seandillon92: Could this be renamed exportInstances? it would be just an overload

Plugins~/Src/MeshSyncClientBlender/msblenGeometryNodesUtils.cpp
schinkowski: It's a different kind of correction.
Lights and cameras need to be rotated and flipped so they face the correct way in Unity. That is what the first correction does.
The second correction undoes the correction that would have been applied the the instanced light/camera to show correctly in Unity so that the inverse of the object's transform matrix multiplied with the instance transform matrix gives us back the world transform of the instance.

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
schinkowski: Other meshes. The issues is that blender internally creates temporary meshes that seem to be copies of other meshes and frees them later. When meshsync tries to access those later, the stored pointer points to freed memory and we get exceptions.

Plugins~/Src/MeshSyncClientBlender/msblenContextGeometryNodes.cpp
schinkowski: Sure, done

Plugins~/Src/MeshSyncClientBlender/msblenGeometryNodeUtils.h
schinkowski: That would not work with the `applyCorrectionIfNeeded` method. That method uses `is_light` from the mesh utils and doesn't check the type directly.

seandillon92: Both the is_camera and is_light are a type check
```
bool is_camera(const Object *obj) { return obj->type == OB_CAMERA; }
bool is_light(const Object *obj) { return obj->type == OB_LAMP; }
```

can we not just pass that short? And the method can just check the parameter instead of using the utility

seandillon92: you can create an overload that takes the enum instead of the object. Then the function that takes object can be simplified to
```
applyCorrectionIfNeeded(obj, ...){
applyCorrectionIfNeeded(obj->type, ..)
}
```

schinkowski: Yes but `applyCorrectionIfNeeded` specifically uses `is_camera` and `is_light` to avoid direct type checks. Unless we don't care about that abstraction anymore and want to get rid of those methods.
I think it would better to just pass the object instead of removing the abstraction just so we can pass a type from the object with no real benefit.

seandillon92 (commented):
I started the PR so github does not let me cannot approve, but it looks good to me

Plugins~/Src/MeshSyncClientBlender/msblenGeometryNodeUtils.h
seandillon92: I have 2 longer term concerns: 
One is passing around raw pointers to blender objects. This is probably necessary for a lot of our code, but if there is another way, it might be good to start thinking about how we can change our code.

The other is, the function takes an object to convert a matrix. I understand why, but it seems to me that this function should do what it describes, and the object parameter looks strange. 

In any case, I approve the PR as this is not blocking

schinkowski:
Need this https://github.com/unity3d-jp/MeshSync/pull/770

sindharta (approved):

seandillon92 (commented):
LGTM

PR #263, opened 2022-09-05
[WIP] Inverse installation - Blender to Unity

seandillon92:
Depends on https://github.com/unity3d-jp/MeshSync/pull/769


seandillon92:
@sindharta a few questions while this is still WIP:
1. Would you prefer if I break it down to 2 PRs? I can create a separate one for the changes in the python scripts
2. The python code is implemented as multiple Operators that call each other. Do you know a better way to do this?

PR #264, opened 2022-09-06
[For 0.15.x] blender-feat: Editor commands in Blender clients

seandillon92:
Introducing callbacks for sending commands to the Editor.
Depends on https://github.com/unity3d-jp/MeshSync/pull/769


Plugins~/Src/MeshSyncClientBlender/MeshSyncClientBlender.cpp
```diff
@@ -107,6 +107,8 @@ PYBIND11_MODULE(MeshSyncClientBlender, m)
 
             BindConst(is_server_available, self->isServerAvailable())
             BindConst(error_message, self->getErrorMessage())
+            BindConst(is_editor_server_available, self->isEditorServerAvailable())
```
sindharta, line 110: Since they are both "editor" servers, I think we should rename this.
Perhaps, `autosetup_server` ?

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
```diff
@@ -1350,6 +1352,20 @@ bool msblenContext::isServerAvailable()
     return m_sender.isServerAvaileble();
 }
 
+bool msblenContext::isEditorServerAvailable()
+{
+    ms::ClientSettings settings = ms::ClientSettings();
+    settings.port = 8081;
```
sindharta: I think we should use the same IP address as the one used for the original server.
Plus, shouldn't we make the port configurable instead of hardcoding it ?

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
```diff
@@ -1408,6 +1424,21 @@ void msblenContext::requestLiveEditMessage()
     m_sender.requestLiveEditMessage();
 }
 
+bool msblenContext::sendEditorCommand(ms::EditorCommandMessage::CommandType type)
+{
+    ms::ClientSettings settings = ms::ClientSettings();
+    settings.port = 8081;
```
sindharta: same as above

Plugins~/Src/MeshSyncClientBlender/MeshSyncClientBlender.cpp
seandillon92: MeshSyncServer can be used at runtime as well, and will only work while the containing scene is active. With that in mind I would call them Scene Servers vs Editor Servers. I think naming the classes AutoSetup/AutoSetupSettings ties their functionality to a very specific purpose, while the server should be just executing commands. Might be better to rename them EditorServer/EditorServerSettings in the other PR and keep these names as they are. What do you think?

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
seandillon92: I will make it configurable from the panel.

Plugins~/Src/MeshSyncClientBlender/MeshSyncClientBlender.cpp
seandillon92: Made the change I mentioned in this PR for now https://github.com/unity3d-jp/MeshSync/pull/774

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
seandillon92: Changed it to take the IP address from client settings and the port from a new field in the BaseSyncSettings that is exposed to the python layer

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
seandillon92: Changed it to take the IP address from client settings and the port from a new field in the BaseSyncSettings that is exposed to the python layer

SonnyCampbellUnity (approved):

sindharta (approved):

PR #265, opened 2022-09-06
Install MeshSync from Blender 1st PR

seandillon92:
Depends on https://github.com/Unity-Technologies/MeshSyncDCCPlugins/pull/264


Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_common.py
```diff
@@ -66,6 +79,54 @@ def msb_on_animation_settings_updated(self = None, context = None):
     # nothing to do for now
     return None
 
+def msb_on_unity_project_path_updated(self = None, context = None):
+    #TODO invoke callback for installing meshsync
+    return None
+
+def msb_get_editor_path_prefix(hub_path = None):
+    prefix = hub_path
+    if prefix is None:
```
sindharta: I have this error when trying the plugin.
image

As I mentioned before, 
1. the actual location of the editors can be configured in the hub
2. the location of the hub itself can also be configured

Is there a plan to tackle the above issues as well ?

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_common.py
```diff
@@ -102,15 +170,250 @@ def draw(self, context):
...
+
+    filepath : bpy.props.StringProperty(name = "filepath")
+
+    def execute(self, context):
+
+        context.scene.meshsync_unity_project_path = self.filepath
+
+        if msb_validate_project_path(self.filepath) == False:
+            message = self.filepath + " is not a Unity project"
+            MS_MessageBox(message = message)
```
sindharta: Right now, when I have an invalid path and I click on "AutoSync" or "ManualSync" button, it shows a folder selection window, instead of outputting this message box.
Is this correct ?

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_common.py
```diff
@@ -102,15 +170,250 @@ def draw(self, context):
...
+class MESHSYNC_OT_InstallMeshSync(bpy.types.Operator):
+
+    bl_idname = "meshsync.install_meshsync"
+    bl_label = "Try to get the unity path from a running server"
+
+    def get_meshsync_path(self):
+        script_file = os.path.realpath(__file__)
+        directory = os.path.dirname(script_file)
+        system = platform.system()
+        if system == 'Windows':
```
sindharta: Aren't we already modifying the manifest so that the project downloads a particular MeshSync version from the Internet ?

If so, why are we bundling a particular version on MeshSync in the plugin ?

This will make the size of MeshSyncDCCPlugins a lot bigger.
Probably around (15 MB * number of supported Blender versions) ?

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_common.py
```diff
@@ -102,15 +170,250 @@ def draw(self, context):
...
+
+    def add_meshsync_to_unity_manifest(self, path, entry):
+        file = open(path, "r+");
+        data = json.load(file);
+
+        # check if MeshSync is installed on selected project
+        # if not installed, install it
+        found = False
+
+        for package in data['dependencies']:
```
sindharta: Checking `package-lock.json` is probably more robust if we want to know whether MeshSync has been added or not

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_common.py
```diff
@@ -102,15 +170,250 @@ def draw(self, context):
...
+            version = version[:-1]
+            return version
+
+    def launch_project(self, editor_path, project_path):
+        os = platform.system()
+        if os == 'Windows':
+            path = editor_path + " -projectPath \"" + project_path + "\""
+        elif os == 'Darwin':
+            path = editor_path + " -projectPath " + project_path
+        elif os == 'Linux':
```
sindharta: The code for Linux and Mac are usually the same, so perhaps we can make the code simpler ?

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_common.py
```diff
@@ -102,15 +170,250 @@ def draw(self, context):
...
+        bpy.ops.meshsync.create_server('INVOKE_DEFAULT')
+        return {'FINISHED'}
+
+class MESHSYNC_OT_CreateServer(bpy.types.Operator):
+    bl_idname = "meshsync.create_server"
+    bl_label = "Create a Scene Server if there is none"
+
+    def execute(self, context):
+       #If there is no server in the scene, add one
+        msb_context.sendEditorCommand(1)
```
sindharta: Is it possible to define a more meaningful variable for this 1 and 2 ?

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_common.py
```diff
@@ -102,15 +170,250 @@ def draw(self, context):
...
     def execute(self, context):
+
+        # Export data
         msb_apply_scene_settings()
-        msb_context.setup(bpy.context);
+        msb_context.setup(bpy.context)
         msb_context.export(msb_context.TARGET_OBJECTS)
-        return{'FINISHED'}
 
+        return{'FINISHED'}
```
sindharta: I believe we have made a flow diagram for this new flow ?
Can we also put it as a picture in our Documentation  ?

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_common.py
seandillon92: > the location of the hub itself can also be configured

This is configurable in the Unity Project panel:
image

> the actual location of the editors can be configured in the hub

This will need to be tackled. A better design might be to have a DIR property for the Editors folder instead of the Hub

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_common.py
seandillon92: Correct, the current design is to prompt the user to select a project location (if not pointing to one), and if that location is not a unity project, it will give that message. Sounds like its confusing, do you think we should just show this message box instead of showing the folder selection dialog?

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_common.py
seandillon92: Good call, I will change this

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_common.py
seandillon92: Sounds good, will change this.

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_common.py
seandillon92: Good idea, I will create a second PR for the docs

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_common.py
seandillon92: Definitely, will do

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_common.py
sindharta: Might be my own preference thing, but yeah, showing this message box would be more intuitive, I guess.

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_common.py
seandillon92: done

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_common.py
seandillon92: done

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_common.py
seandillon92: done, currently set to 0.14.5-preview but will need to be updated with the next version that will contain the server changes

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_common.py
seandillon92: I have made a change to skip the prompt, it started getting confusing for me as well

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_common.py
seandillon92: Changed to point to the editors folder.

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_common.py
```diff
@@ -102,6 +146,226 @@ def draw(self, context):
...
+        for package in data['dependencies']:
+            if (package == 'com.unity.meshsync'):
+                found = True
+
+        if(found == False):
+            #install for user
+            dependencies = data["dependencies"];
+            dependencies["com.unity.meshsync"] = entry
+            if (entry == ""):
+                del dependencies["com.unity.meshsync"]
```
SonnyCampbellUnity: I don't understand this piece. Can you just check if `entry != ""` and only add it to the dependencies then, rather than adding and removing?

seandillon92: I cannot remember why I wrote it this way, but this is dead code, as we will never invoke this with an empty entry. I removed check altogether. Good call!

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_common.py
seandillon92: done

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_installation.py
```diff
@@ -0,0 +1,204 @@
...
+            lock_path = directory + "Packages/packages-lock.json"
+            manifest_path = directory + "Packages/manifest.json"
+        else:
+            lock_path = directory + "/Packages/packages-lock.json"
+            manifest_path = directory + "/Packages/manifest.json"
+
+        #TODO replace with release that has the Editor Commands changes
+        manifest_entry = "0.14.5-preview"
+        if DEBUG == 1:
+            manifest_entry = "file:C:\\Users\\Sean Dillon\\MeshSync\\MeshSync~\\Packages\\com.unity.meshsync"
```
SonnyCampbellUnity: Probably meant to delete this debug code?

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_installation.py
```diff
@@ -0,0 +1,204 @@
...
+            return 'ALREADY_STARTED'
+
+    editor_version = msb_get_editor_version(directory)
+    editor_path = msb_get_editor_path(editor_version)
+
+    if not path.exists(editor_path):
+        return 'EDITOR_NOT_EXISTS'
+
+    #Launch the editor with the project
+    unity_process = msb_launch_project(editor_path, directory)
```
SonnyCampbellUnity: Could this process launch fail? Is there a result or bool or anything you can check here to make sure the process successfully started?

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_installation.py
seandillon92: This is necessary for testing. I will keep this comment unresolved until we make the change

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_installation.py
seandillon92: > Could this process launch fail?

We have checked that the path to the Editor .exe exists before getting here. How would the launch fail? More importantly, how are we supposed to handle it?

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_common.py
seandillon92: Implemented here https://github.com/Unity-Technologies/MeshSyncDCCPlugins/pull/266

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_installation.py
SonnyCampbellUnity: I'm not sure, I just mean if the application failed to start up for some reason (maybe it needs a permission or something - just thinking of possibilities where the process can't start the exe). We should display something to the user saying "Failed to launch editor" so they know they manually need to go and open Unity

SonnyCampbellUnity: It may not be possible, I was just guessing that `subprocess.Popen()` might return a success code or some indication that the process started/failed

seandillon92: > might return a success code or some indication that the process started/failed

Turns out it throws an exception if something goes wrong. I tested by modifying the path to something that does not exists before calling it. The idea to notify the user sounds good to me, better than getting the error stack. I have now implemented it

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_installation.py
```diff
@@ -0,0 +1,251 @@
...
+    return False
+
+def msb_get_editor_path_suffix():
+    os = platform.system()
+    if os == 'Windows':
+        return path.join("Editor","Unity.exe")
+    elif os == 'Darwin':
+        return path.join("Unity.app","Contents","MacOS","Unity")
+    elif os == 'Linux':
+        return path.join("Editor","Unity")
```
seandillon92, line 48: the doc about the location of the Editor seems wrong https://docs.unity3d.com/Manual/EditorCommandLineArguments.html
I tested on Ubuntu, its under Editor/Unity

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_installation.py
```diff
@@ -0,0 +1,228 @@
...
+        self.layout.label(text=message)
+    bpy.context.window_manager.popup_menu(draw, title = title, icon = icon)
+
+def msb_get_meshsync_entry():
+    if not DEBUG:
+        return "0.14.5-preview"
+
+    os = platform.system()
+    if os == 'Windows':
+        return "file:C:\\Users\\Sean Dillon\\MeshSync\\MeshSync~\\Packages\\com.unity.meshsync"
```
sindharta: Sean Dillon ?

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_installation.py
```diff
@@ -0,0 +1,228 @@
...
+    elif os == 'Darwin' or os == 'Linux':
+        return "file:/Users/sean.dillon/MeshSync/MeshSync~/Packages/com.unity.meshsync"
```
sindharta: This one too

seandillon92: done

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_installation.py
seandillon92: this is removed now

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_installation.py
seandillon92: Removed the code

sindharta (approved):

PR #266, opened 2022-09-20
blender-doc: add doc for installing MeshSync via Blender

seandillon92:


Documentation~/Installation.md
```diff
@@ -1,6 +1,18 @@
 # Installation
 
-## Automatic Installation
+## Automatic Installation via DCC tools
+We can install everything through DCC tools. Currently, this is available only for Blender. For other DCC tools, see [Automatic Installation via Unity Editor](./Installation.md#automatic-installation-via-unity-editor).
```
sindharta: Can you please elaborate here what you mean by "everything" here ?

Documentation~/Installation.md
```diff
@@ -1,6 +1,18 @@
...
+### Blender
+1. Download the Blender add-on from [here](https://github.com/Unity-Technologies/MeshSyncDCCPlugins/tree/installation/blender-docs/Editor/Plugins~).
```
sindharta: We should probably point the link to https://github.com/Unity-Technologies/MeshSyncDCCPlugins/releases

Documentation~/Installation.md
```diff
@@ -1,6 +1,18 @@
 # Installation
 
-## Automatic Installation
+## Automatic Installation via DCC tools
```
sindharta: Perhaps "Unity Project Setup" via DCC tools is a better naming here ?

Documentation~/Installation.md
seandillon92: Trying to convey both packages, MeshSync and the add-on zip. Might be better to phrase it as : "We can install MeshSync through DCC tools"?

Documentation~/Installation.md
seandillon92: done

Documentation~/Installation.md
seandillon92: done

Documentation~/Installation.md
seandillon92: I made a few changes in general that should be more clear, let me know what you think

Documentation~/Installation.md
```diff
@@ -1,6 +1,23 @@
...
 
-## Automatic Installation
+## Unity Project Setup via DCC tools
+We can install MeshSync on a Unity project through DCC tools. This install method is designed for users who are not familiar with the Unity Editor or the Unity Package Manager.
+
+Currently, this install method is available only for Blender. For other DCC tools, see [MeshSync](https://docs.unity3d.com/Packages/com.unity.meshsync@latest)'s installation page.
+### Blender
+Install the Blender add-on:
+1. Download the latest MeshSyncDCCPlugins release from [here](https://github.com/Unity-Technologies/MeshSyncDCCPlugins/releases).
+1. Install the add-on manually, as described [here](./Installation.md#blender-1).
```
sindharta, line 10: Might be better if we are telling the users to reuse the zip files from step 1 directly instead of telling them open `Packages/MeshSync DCC Plugins/Editor/Plugins~` which might be confusing.

Approving this for now though.

sindharta (approved):

PR #275, opened 2022-09-29
blender-fix: fix for crash in blender 3.3

schinkowski:


Plugins~/Src/MeshSyncClientBlender/msblenBinder.cpp
```diff
@@ -375,7 +375,12 @@ bool blender::BObject::is_selected() const
 
 Mesh* BObject::to_mesh() const
 {
+#if BLENDER_VERSION >= 303
+    // In blender 3.3 the pointer to interpret the bool moves by 8 instead of 1 for some reason so this needs to be a type of that size:
```
seandillon92, line 379: Is this change in Blender 3.3 specific to to_mesh or is it affecting other parts as well?

sindharta (approved):

sindharta:
Is this fix for crashes that happen when syncing meshes in general (didn't happen to me though) or when syncing geometry nodes only ?

Plugins~/Src/MeshSyncClientBlender/msblenBinder.cpp
schinkowski: I only found the issue in this call.

sindharta:
@seandillon92 Can you review this code and give approval if it looks ok ?

seandillon92 (approved):
Looks good to me

PR #276, opened 2022-09-29
blender: use default ports if possible in auto-config

seandillon92:


Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_installation.py
```diff
@@ -213,6 +214,26 @@ def msb_try_start_unity_project (context, directory):
 
     return 'UNKNOWN'
 
+def msb_is_port_available(port):
+    try:
+        with closing(socket.socket(socket.AF_INET, socket.SOCK_STREAM)) as s:
+            s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
```
SonnyCampbellUnity: I don't fully understand this, are you opening two sockets as `s` here?

sindharta (approved):

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_installation.py
seandillon92: yes, but its an error, will remove the line

SonnyCampbellUnity (approved):

PR #288, opened 2022-10-10
feat: Material support

schinkowski:


Plugins~/Src/MeshSyncClientBlender/msblenMaterialsExportHelper.cpp
```diff
@@ -0,0 +1,429 @@
...
+		}
+
+		if (!m_settings->sync_textures) {
+			setTextureHandler = nullptr;
+		}
+
+		switch (sourceNode->type) {
+		case SH_NODE_TEX_IMAGE:
+		{
+			if (sourceNode->id) {
```
SonnyCampbellUnity: Can you flatten out these nested if statements at all by changing to `if(!condition) {break;}` or something like that? It would make this function much easier to read

SonnyCampbellUnity: Would also help if it was possible to separate into functions for each of the individual case statements

Plugins~/Src/MeshSyncClientBlender/msblenMaterialsExportHelper.cpp
```diff
@@ -0,0 +1,429 @@
...
+	}
+}
+
+void msblenMaterialsExportHelper::exportMaterialFromNodeTree(const Material* mat, ms::StandardMaterial& stdmat)
+{
+	bNode* bsdfNode;
+	bNode* outputNode;
+
+	if (!getBSDFAndOutput(mat, bsdfNode, outputNode)) {
+		return;
```
SonnyCampbellUnity, line 473: Is there any indication about what's wrong to the user if the material sync doesn't work ?

Plugins~/Src/MeshSyncClientBlender/msblenMaterialsExportHelper.cpp
schinkowski: I've extracted it to methods, returning early isn't really possible with these, it does different logic depending what's connected or not.

Plugins~/Src/MeshSyncClientBlender/msblenMaterialsExportHelper.cpp
schinkowski: This can only return false if there is no material output node or no bsdf connected to the material output. 
Every material should have these or it would just be black.

Plugins~/Src/MeshSyncClient/Include/MeshSyncClient/msTextureManager.h
```diff
@@ -24,6 +24,8 @@ class TextureManager
     int addFile(const std::string& path, TextureType type);
     // thread safe
     int add(TexturePtr tex);
+    // thread safe
+    int addPackedImage(const std::string& name, const void* data, size_t size, TextureType type);
```
seandillon92: What is a packed image?

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
```diff
@@ -25,6 +26,9 @@
 #include "DNA_node_types.h"
 #include <sstream>
 #include "msblenBinder.h"
+#include "BlenderPyObjects/BlenderPyID.h"
+#include "BlenderPyObjects/BlenderPyID.h"
+#include "BlenderPyObjects/BlenderPyID.h"
```
seandillon92: This is included thrice?

Plugins~/Src/MeshSyncClientBlender/msblenMaterialsExportHelper.cpp
```diff
@@ -0,0 +1,472 @@
...
+					if (STREQ(inputSocket->name, shaderIdentifier)) {
+						bNode* connectedBSDF = removeReroutes(getNodeConnectedToSocket(inputSocket), mat);
+						if (connectedBSDF)
+						{
+							bsdf = connectedBSDF;
+							break;
+						}
+					}
+				}
+			}
```
seandillon92: to avoid deep nesting, is it possible to extract this into a callback that returns the bsdf?

Plugins~/Src/MeshSyncClientBlender/msblenMaterialsExportHelper.cpp
```diff
@@ -0,0 +1,472 @@
...
+	return m_texture_manager->addFile(path, type);
+}
+
+void msblenMaterialsExportHelper::handleImageNode(ms::TextureType& textureType,
+	bool resetIfInputIsTexture,
+	std::function<void(const mu::float4& colorValue)> setColorHandler,
+	std::function<void(int textureId)> setTextureHandler,
+	bNode* sourceNode)
+{
+	if (sourceNode->id) {
```
seandillon92: is there any way to reduce nesting in this function? I find it difficult to follow

Plugins~/Src/MeshSyncClientBlender/msblenMaterialsExportHelper.cpp
```diff
@@ -0,0 +1,472 @@
...
+
+void msblenMaterialsExportHelper::setValueFromSocket(const Material* mat,
+	bNodeSocket* socket,
+	ms::TextureType textureType,
+	bool resetIfInputIsTexture,
+	std::function<void(const mu::float4& colorValue)> setColorHandler,
+	std::function<void(int textureId)> setTextureHandler)
+{
+	// If there is an image linked to the socket, send that as a texture:
+	if (socket->link) {
```
seandillon92: Can you check the negation and return after handling the socket value? Should help reducing nesting

Plugins~/Src/MeshSyncClientBlender/msblenMaterialsExportHelper.cpp
```diff
@@ -0,0 +1,472 @@
...
+	default:
+		stdmat.setShader("Default");
+		break;
+	}
+}
+
+void msblenMaterialsExportHelper::setHeightFromOutputNode(const Material* mat, ms::StandardMaterial& stdmat, bNode* outputNode)
+{
+	auto displacementSocket = getInputSocket(outputNode, displacementIdentifier);
+	if (displacementSocket)
```
seandillon92: can you return on negation here?

Plugins~/Src/MeshSyncClient/Include/MeshSyncClient/msTextureManager.h
schinkowski: An image that's packed into the blend file.

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
schinkowski: No idea, fixed

Plugins~/Src/MeshSyncClientBlender/msblenMaterialsExportHelper.cpp
schinkowski: Done

Plugins~/Src/MeshSyncClientBlender/msblenMaterialsExportHelper.cpp
schinkowski: I've created a method. Not sure how useful it is to have so many methods for every little piece of functionality that's only used once though. It can be harder to read than having it all in a single place.

Plugins~/Src/MeshSyncClientBlender/msblenMaterialsExportHelper.cpp
schinkowski: Done

Plugins~/Src/MeshSyncClientBlender/msblenMaterialsExportHelper.cpp
schinkowski: Done

Plugins~/Src/MeshSyncClient/Include/MeshSyncClient/msTextureManager.h
seandillon92: This file is in the MeshSyncClient namespace, which is independent of Blender or any DCC tool. Can this be renamed to addImage, so that it leaves out any Blender specific terms?

Plugins~/Src/MeshSyncClientBlender/msblenMaterialsExportHelper.cpp
```diff
@@ -0,0 +1,486 @@
...
+const auto heightIdentifier = "Height";
+const auto scaleIdentifier = "Scale";
+
+bNode* removeReroutes(bNode* node, const Material* mat) {
+	if (!node) {
+		return nullptr;
+	}
+
+	auto tree = mat->nodetree;
+	if (node->type == NODE_REROUTE)
```
seandillon92: Can you return node on negation here? will help reduce nesting a bit more

Plugins~/Src/MeshSyncClientBlender/msblenMaterialsExportHelper.cpp
```diff
@@ -0,0 +1,486 @@
...
+bNode* getNodeConnectedToSocket(bNodeSocket* socket) {
+	if (socket->link) {
+		return socket->link->fromnode;
+	}
+
+	return nullptr;
+}
+
+bNode* handlePassthroughBSDF(const Material* mat, bNode* bsdf) {
+	if (bsdf && bsdf->type == SH_NODE_MIX_SHADER) {
```
seandillon92: similarly here, can you return bsdf if the argument is null or its not a mix shader?

Plugins~/Src/MeshSyncClientBlender/msblenMaterialsExportHelper.cpp
```diff
@@ -0,0 +1,486 @@
...
+
+bNode* getNodeConnectedToSocket(bNodeSocket* socket) {
+	if (socket->link) {
+		return socket->link->fromnode;
+	}
+
+	return nullptr;
+}
+
+bNode* handlePassthroughBSDF(const Material* mat, bNode* bsdf) {
```
seandillon92: Seems like this function handles the mix shader node. Does it make sense to be called handleMixShader or similar?

Plugins~/Src/MeshSyncClientBlender/msblenMaterialsExportHelper.cpp
```diff
@@ -0,0 +1,486 @@
...
+const auto emissionIdentifier = "Emission";
+const auto emissionStrengthIdentifier = "Emission Strength";
+
+const auto shaderIdentifier = "Shader";
+
+const auto displacementIdentifier = "Displacement";
+const auto heightIdentifier = "Height";
+const auto scaleIdentifier = "Scale";
+
+bNode* removeReroutes(bNode* node, const Material* mat) {
```
seandillon92: What does this function do exactly?

Plugins~/Src/MeshSyncClient/Include/MeshSyncClient/msTextureManager.h
schinkowski: I've refactored it to use the other method and removed this one now.

Plugins~/Src/MeshSyncClientBlender/msblenMaterialsExportHelper.cpp
schinkowski: Done

Plugins~/Src/MeshSyncClientBlender/msblenMaterialsExportHelper.cpp
schinkowski: Done

Plugins~/Src/MeshSyncClientBlender/msblenMaterialsExportHelper.cpp
schinkowski: Done, it was supposed to become a more generic method to handle more in the future but we can change that when needed, I guess.

Plugins~/Src/MeshSyncClientBlender/msblenMaterialsExportHelper.cpp
schinkowski: It removes reroute nodes and returns the actual input node upstream.

seandillon92: I would be useful to have some doc in our confluence, describing the issues with different types of nodes and how we handle them. It's difficult to get what the code tries to do without some context

Plugins~/Src/MeshSyncClientBlender/msblenMaterialsExportHelper.cpp
```diff
@@ -0,0 +1,484 @@
...
+
+	return nullptr;
+}
+
+int msblenMaterialsExportHelper::exportTexture(const std::string& path, ms::TextureType type) const
+{
+	return m_texture_manager->addFile(path, type);
+}
+
+void msblenMaterialsExportHelper::handleImageNodeWithAssignedImage(ms::TextureType& textureType,
```
seandillon92: This function does 2 things: 
1. invokes the texture handler on the texture, which might be either in the blendfile or somewhere in the filesystem
2. invokes the color handler

Can these 2 responsibilities be broken down into 2 smaller functions? It should allow us to reduce some nesting by having some guard checks at the top with early exits

seandillon92: Or probably even better, extract the code that exports an image from a blend file to it's own function

Plugins~/Src/MeshSyncClientBlender/msblenMaterialsExportHelper.cpp
```diff
@@ -0,0 +1,484 @@
...
+	case SH_NODE_GAMMA:
+	case SH_NODE_HUE_SAT:
+	{
+		handlePassthrough(mat, textureType, resetIfInputIsTexture, setColorHandler, setTextureHandler, sourceNode);
+		break;
+	}
+	}
+}
+
+void msblenMaterialsExportHelper::setShaderFromBSDF(ms::StandardMaterial& stdmat, bNode* bsdfNode)
```
seandillon92, line 351: Is the standard material referring to the shader in Unity? Since we support multiple pipelines, the name can be misleading. Would something like PBRMaterial make more sense?

Plugins~/Src/MeshSyncClientBlender/msblenMaterialsExportHelper.cpp
schinkowski: Do we need it on confluence or would the user documentation suffice? I'm working on the user docs for this.

schinkowski: I've added some here https://github.com/Unity-Technologies/MeshSyncDCCPlugins/blob/material_support/Documentation~/Blender.md#Material-sync-mode

Plugins~/Src/MeshSyncClientBlender/msblenMaterialsExportHelper.cpp
schinkowski: This is what it was called in meshsync before. It refers to Unity's standard shader.

Plugins~/Src/MeshSyncClientBlender/msblenMaterialsExportHelper.cpp
schinkowski: Done

Plugins~/Src/MeshSyncClientBlender/msblenMaterialsExportHelper.cpp
seandillon92: I see that the 3dsmax client is using this class as well. What if we used a new class ,i.e. PBRSurfaceMaterial that has the same functionality as the standard material and call the StandardMaterial obsolete? If we were to follow the pattern that exists in the codebase, we would have to define a new class for each shader or RP. And that would also assume we pack the channels for HDRP in the DCC tool.

Plugins~/Src/MeshSyncClientBlender/msblenMaterialsExportHelper.cpp
```diff
@@ -0,0 +1,497 @@
...
+		if (extensionIndex > 0) {
+			std::string extension = filepath.substr(extensionIndex);
+
+			if (name.find(extension) != name.length() - extension.length()) {
+				name += extension;
+			}
+		}
+
+		int exported = m_texture_manager->addImage(name, 0, 0, imagePackedFile->packedfile->data, imagePackedFile->packedfile->size, ms::TextureFormat::RawFile, textureType);
+		setTextureHandler(exported);
```
seandillon92, line 136: To keep this function from doing more than one thing, I think this function should just export images, this handler should be invoked on the caller

Plugins~/Src/MeshSyncClientBlender/msblenMaterialsExportHelper.cpp
schinkowski: My changes were additions to the existing class, other DCC tools that used this would not be affected and we can implement the additional properties on them too, no need to make a new class, I think.
Baking the maps on the DCC tool side would require knowledge about the pipeline of the server and separate implementations for each DCC tool, whereas baking them in Unity is done in a single place for its active pipeline, no matter what DCC tool the data came from so would have a lower maintenance cost.

Plugins~/Src/MeshSyncClientBlender/msblenMaterialsExportHelper.cpp
schinkowski: It can be called multiple times, depending how many images are packed. We'd need to pass in a callback anyway to do that.

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
```diff
@@ -25,6 +26,7 @@
 #include "DNA_node_types.h"
 #include <sstream>
 #include "msblenBinder.h"
+#include "BlenderPyObjects/BlenderPyID.h"
```
SonnyCampbellUnity: is this needed?

Plugins~/Src/MeshSyncClientBlender/msblenMaterialsExportHelper.cpp
```diff
@@ -0,0 +1,497 @@
...
+{
+	return m_texture_manager->addFile(path, type);
+}
+
+void msblenMaterialsExportHelper::exportPackedImages(ms::TextureType& textureType,
+	std::function<void(int textureId)> setTextureHandler,
+	Image* img) const
+{
+	for (auto imagePackedFile : list_range((ImagePackedFile*)img->packedfiles.first)) {
+		std::string name = img->id.name + 2; // +2 because of IM prefix!
```
SonnyCampbellUnity: What sort of black magic is this? 

Does this create a string in the variable `name` that is equal to `img->id.name` but two empty characters longer?

SonnyCampbellUnity: Or is this doing some pointer arithmetic from the start of the `img->id.name` character array?

Plugins~/Src/MeshSyncClientBlender/msblenMaterialsExportHelper.cpp
```diff
@@ -0,0 +1,497 @@
...
+		int exported = m_texture_manager->addImage(name, 0, 0, imagePackedFile->packedfile->data, imagePackedFile->packedfile->size, ms::TextureFormat::RawFile, textureType);
+		setTextureHandler(exported);
+	}
+}
+
+void msblenMaterialsExportHelper::exportImageFromImageNode(ms::TextureType& textureType,
+	std::function<void(int textureId)> setTextureHandler,
+	bNode* sourceNode) const
+{
+	if (setTextureHandler) {
```
SonnyCampbellUnity: Can you switch this to `if (!setTextureHandler){return;}` to reduce the indentation

Plugins~/Src/MeshSyncClientBlender/msblenMaterialsExportHelper.cpp
```diff
@@ -0,0 +1,497 @@
...
+			setValueFromSocket(mat,
+				imageInput, textureType,
+				resetIfInputIsTexture,
+				nullptr,
+				setTextureHandler);
+		}
+	}
+
+	auto strengthInput = getInputSocket(sourceNode, normalStrengthIdentifier);
+	if (strengthInput && setColorHandler) {
```
SonnyCampbellUnity, line 216: Can you do this same sort of logic on the other if statement in the function to be consistent & reduce nesting?

Plugins~/Src/MeshSyncClientBlender/msblenMaterialsExportHelper.cpp
SonnyCampbellUnity: If that StandardMaterial is also being used for URP/HDRP, maybe we should rename it to something more generic, or create a new one as Sean mentioned, because it won't just be used for Unity's standard shader anymore

Plugins~/Src/MeshSyncClientBlender/msblenMaterialsExportHelper.cpp
SonnyCampbellUnity: It would still be useful to let the user know exactly what's wrong if there's a material being synced with no output node, or no bsdf connected to the output node

SonnyCampbellUnity: And even in general, if the material sync fails, do we let the user know if there are nodes or things we can't handle?

Plugins~/Src/MeshSyncClientBlender/msblenMaterialsExportHelper.cpp
seandillon92: > other DCC tools that used this would not be affected
 
 This is not a concern for the work you have done, but it should be for renaming the class, as I suggest
 

>  Baking the maps on the DCC tool side would require knowledge about the pipeline of the server and separate implementations for each DCC tool, whereas baking them in Unity is done in a single place for its active pipeline, no matter what DCC tool the data came from so would have a lower maintenance cost.

I agree, this is why we must not have classes on the client that are specific to the shader used in Unity

seandillon92: > maybe we should rename it

We have to be careful here to not break the 3dsmax code. This is why I propose calling the current class obsolete and creating a new one

Plugins~/Src/MeshSyncClientBlender/msblenMaterialsExportHelper.cpp
seandillon92: Ah ok, I missed that

Plugins~/Src/MeshSyncClientBlender/msblenMaterialsExportHelper.cpp
seandillon92: I have used the same pattern before, its pointer arithmetic as you mentioned.

Plugins~/Src/MeshSyncClientBlender/msblenMaterialsExportHelper.cpp
schinkowski: Would we want to show that kind of message in blender or Unity?

Plugins~/Src/MeshSyncClientBlender/msblenMaterialsExportHelper.cpp
seandillon92: I think we also need a design doc

Plugins~/Src/MeshSyncClientBlender/msblenMaterialsExportHelper.cpp
mloccy: Recommend taking another stab at this as this is a bit dangerous as it is currently written, right now it's assuming the string is at least 2 characters long when there's no guarantee that that's always the case, so putting a length check in would be prudent to avoid a possible segfault in that scenario.

Recommend also avoiding the pointer arithmetic simply because its kinda gross and I was raised to abhor it, if you choose to do that std::string_view is your friend as it allows you to use a c-style string like a std::string (with an iterator) and it has the added bonus of not copying anywhere, which in this scenario is a micro-optimization so not too important. You can read more about that here https://en.cppreference.com/w/cpp/string/basic_string_view

Plugins~/Src/MeshSyncClientBlender/msblenMaterialsExportHelper.cpp
SonnyCampbellUnity: I suppose we wouldn't really have knowledge of why the material sync failed in Unity unless we sent over a specific failure message to display there...

schinkowski: Yeah that's what I'm wondering. We can implement sending text messages to Unity or just show the error in blender. In Unity we could log it as an error, in blender there's no unobtrusive way but still obvious way to show it.

Plugins~/Src/MeshSyncClientBlender/msblenMaterialsExportHelper.cpp
schinkowski: I still don't see the value in renaming this. It just means it's a standard material with a standard pbr shader. If in the future we support special materials or shader effects these would be new classes with different maps and shaders.

Plugins~/Src/MeshSyncClientBlender/msblenMaterialsExportHelper.cpp
schinkowski: Is the lucid chart enough? I made a confluence link to it here https://confluence.unity3d.com/display/ARTDEV/Material+sync+mode

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
schinkowski: Not anymore, deleted now.

Plugins~/Src/MeshSyncClientBlender/msblenMaterialsExportHelper.cpp
seandillon92: In this case, we should have a good xml comment clarifying that this is not referring to the Unity Standard Shader, its just a generic PBR material

Plugins~/Src/MeshSyncClientBlender/msblenMaterialsExportHelper.cpp
schinkowski: Yes, blender has 2 character ID prefixes in front of every name of an object that need to be skipped.
I've refactored it to use a string_view now.

Plugins~/Src/MeshSyncClientBlender/msblenMaterialsExportHelper.cpp
schinkowski: Done

Plugins~/Src/MeshSyncClientBlender/msblenMaterialsExportHelper.cpp
schinkowski: Do it where? Not sure I understand.

Plugins~/Src/MeshSyncClientBlender/msblenMaterialsExportHelper.cpp
schinkowski: I've updated the comment at the top of the StandardMaterial class.

Plugins~/Src/MeshSyncClientBlender/msblenMaterialsExportHelper.cpp
schinkowski: I've added error logging functionality on a separate branch that can be merged after this. Trying not to add more features to this PR.

sindharta:
Files under the `Tests` folder are supposed to be used for CI package tests.
For test assets, I would recommend to put them in another folder, such as `TestAssets` or `Plugins~/TestAssets`

seandillon92 (approved):

Documentation~/Blender.md
```diff
@@ -69,3 +70,36 @@ This option will only appear if we install [ProBuilder](https://docs.unity3d.com
 > If **Bake Modifiers** option is enabled in the plugin, 
 > a mesh that is sent from Unity back into Blender will have modifiers applied beforehand and have them removed after.
 
+## Material sync mode
+
+The MeshSyncServer in Unity always creates a material with a default shader for the active render pipeline.
+There are multiple options to export materials from blender:
```
sindharta: blender with big B.

Documentation~/Blender.md
```diff
@@ -69,3 +70,36 @@ This option will only appear if we install [ProBuilder](https://docs.unity3d.com
...
+
+### None
```
sindharta: Please put this None and basic into a table. 
See [this for reference](https://github.com/Unity-Technologies/MeshSyncDCCPlugins/blob/fe9abd792916acaa87049ada5e3114f8995a4d4c/Documentation~/Blender.md?plain=1#L12).

Because the basic option is a bit long, we can write a summary in the table with an anchor to another section.

Documentation~/Blender.md
```diff
@@ -69,3 +70,36 @@ This option will only appear if we install [ProBuilder](https://docs.unity3d.com
...
+
+Only the material names are exported to Unity and default materials created. Users can override the materials in the mapping on the MeshSyncServer to use their own materials.
+
+### Basic
+
+The materials are synced to Unity in a limited way.
+MeshSync looks for an active material output node and exports the BSDF connected to that.
```
sindharta: The explanation of the basic option is not clear enough in my opinion because there are a lot of ifs.

Some ideas that come to my mind.
1. Clearly separate what happens if a BSDF is found or not 
2. Put and describe what happens to each socket in a table 
3. Clearly put what is supported, what is partially supported, and what is not supported. Something like [this](https://github.com/Unity-Technologies/MeshSyncDCCPlugins/blob/fe9abd792916acaa87049ada5e3114f8995a4d4c/Documentation~/index.md?plain=1#L9)

Tests/Test scenes.meta
```diff
@@ -0,0 +1,8 @@
+fileFormatVersion: 2
```
sindharta: I wrote about the Test scene [here](https://github.com/Unity-Technologies/MeshSyncDCCPlugins/pull/288#issuecomment-1292133617) before.

Copy pasting it here.

Files under the `Tests` folder are supposed to be used for CI package tests.
For test assets, I would recommend to put them in another folder, such as `TestAssets` or `Plugins~/TestAssets`

Plugins~/Src/MeshSyncClientBlender/msblenMaterialsExportHelper.cpp
SonnyCampbellUnity:
```
auto imageInput = getInputSocket(sourceNode, colorIdentifier);
if (setTextureHandler && imageInput) {
        setValueFromSocket(mat,
				imageInput, textureType,
				resetIfInputIsTexture,
				nullptr,
				setTextureHandler);	
}
```

Documentation~/Blender.md
schinkowski: Done

Documentation~/Blender.md
schinkowski: Done

Documentation~/Blender.md
schinkowski: Done

Tests/Test scenes.meta
schinkowski: Done

Plugins~/Src/MeshSyncClientBlender/msblenMaterialsExportHelper.cpp
schinkowski: Ah ok, done

Documentation~/Blender.md
```diff
@@ -69,3 +70,27 @@ This option will only appear if we install [ProBuilder](https://docs.unity3d.com
...
+
+|**Mode** |**Description** |
+|:---       |:---|
+| None | Only the material names are exported to Unity and default materials created. Users can override the materials in the mapping on the MeshSyncServer to use their own materials. |
+| Basic | The materials are synced to Unity in a limited way. Refer to [Basic Material sync mode](#Basic-Material-sync-mode) for more details.||
+
+## Basic Material sync mode
+MeshSync looks for an active material output node and exports the BSDF connected to that. 
+Colors and textures assigned to the active BSDF are exported.
+Smoothness and metallic are baked into maps required by Unity depending on the active render pipeline.
```
sindharta: Nice rewrite. 
I think it may be better to explicitly mention that we support built-in / URP / and HDRP materials.

schinkowski: Done

SonnyCampbellUnity (approved):

seandillon92 (approved):

sindharta (approved):

PR #289, opened 2022-10-14
blender-fix: fix for broken uvs on edit mode

seandillon92:
The root cause seems to be that we are trying to extract UVs from the mesh while indices are extracted from the edit mesh. I noticed that the indices are different in number between the mesh and the edit mesh, and so I don't think we can map the uv coordinates between them, reliably. 

From the file history, the old solution was originally replaced to support multiple UV channels. I am not sure how we can support this now, but at least the UVs should be correct in the one channel we support. We should also call out this caveat in our docs. I can implement the docs change after we discuss this solution.

Potential fix for some of the issues reported here: https://github.com/unity3d-jp/MeshSync/issues/768


sindharta:
Lets talk about this in our next meeting

sindharta (approved):

PR #290, opened 2022-10-21
blender-fix: More reliable way to check if project is already running

seandillon92:


SonnyCampbellUnity (approved):

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_installation.py
```diff
@@ -17,6 +17,26 @@
...
+            os = platform.system()
+            if os == "Darwin" or os == "Linux":
+                import fcntl
+                fcntl.lockf(file.fileno(), fcntl.LOCK_EX|fcntl.LOCK_NB)
+                fcntl.lockf(file.fileno(), fcntl.LOCK_UN)
+            elif os == "Windows":
+                import msvcrt
+                msvcrt.locking(file.fileno(), msvcrt.LK_NBLCK, 0)
+                msvcrt.locking(file.fileno(), msvcrt.LK_UNLCK, 0)
+    except:
```
mloccy: do you think it would be possible to catch more specific exceptions to those api calls?

seandillon92: Sure, will look into it

sindharta (approved):

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_installation.py
seandillon92: @mloccy I was looking at the exceptions by fcntl msvcrt and open. I cannot find any exception that would make sense to not swallow in this example. Do you know any scenario where we would benefit from specifying the exceptions we handle?

seandillon92: I changed this to handle only OS errors. It seems to catch errors correctly on win and macOS (Permission Error and Resource temporarily Unavailable) when trying to lock the file.

PR #293, opened 2022-10-22
blender-fix: get latest compatible version of MeshSync package in AutoSetup

seandillon92:


sindharta (approved):

SonnyCampbellUnity (approved):

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_installation.py
```diff
@@ -16,19 +19,73 @@
...
-    return "0.15.0-preview"
+
+    global meshsync_version
+    if meshsync_version is not None:
+        return meshsync_version
+
+    ssl._create_default_https_context = ssl._create_unverified_context
+    
+    page = 1
+    while True:
```
schinkowski: Why is this in a loop? What are the chances of this becoming an infinite loop?

seandillon92: because we have to get results in pages. We exit the loop when the response payload is empty. We could set a max, perhaps to 100 iterations to be safe

schinkowski: Yeah that might be good, just in case

seandillon92: Done

PR #297, opened 2022-11-11
blender: use hub to select or create project

seandillon92:
This PR does not include giving better feedback to the user through the layout.

### DIFF
--- a/Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_preferences.py
+++ b/Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_preferences.py
@@ -6,6 +6,8 @@ from .unity_mesh_sync_installation import *
 from . import MeshSyncClientBlender as ms
 from bpy.types import AddonPreferences
 from bpy.props import StringProperty, IntProperty, BoolProperty
+from threading import Thread
+from time import sleep
 
 msb_context = ms.Context()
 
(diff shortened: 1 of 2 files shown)


Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_preferences.py
```diff
@@ -18,6 +20,73 @@ def msb_get_editor_path_prefix_default():
...
+
+    @classmethod
+    def description(cls, context, properties):
+        return "Open Unity Hub to select a Unity project to sync data to"
+
+    def invoke(self, context, event):
+
+        if MESHSYNC_OT_OpenHub.active:
+            MESHSYNC_OT_OpenHub.active = False
+            if self.logs_thread is not None:
```
mloccy, line 39: Might be worth reading the python docs about using python threads to ensure you get the behavior you need: https://docs.python.org/3/library/threading.html. From reading it seems to suggest these logs_threads should possible be marked as daemons, and futhermore joining without a minimum timeout might cause this code to hang indefinitely, which could happen in the scenario where log_file.readlines() unexpectedly hangs. A pattern I've seen work is a generous timeout period then an abort where the thread has failed to stop in that time. Might be overly cautious, depending on whether or not this code hanging could hang the blender main thread.

seandillon92: I made some improvements based on the points you raised. 
1. The threads are now marked as daemons.
2. added a timeout of 1s for joining threads
3. replaced readlines with a readline loop. I check the exit condition in the loop to avoid long hangs with readlines.
4. I registered a shutdown function to the atexit event. The function will attempt to join the thread(with a 1s timeout) when the program exits.

schinkowski (approved):

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_preferences.py
```diff
@@ -45,17 +126,32 @@ def editors_path_exists(self):
...
         path = msb_get_hub_path()
-        return dir is not None and os.path.exists(path)
+        return os.path.exists(path)
+
+    def is_hub_supported():
+        hub_version = msb_get_hub_version()
+        if hub_version == "":
+            return False
+
+        return msb_get_most_recent_version(hub_version, "3.0.0") == hub_version
```
SonnyCampbellUnity, line 136: Does this check if the current hub version is 3.0.0? What happens if we're at a higher patch level?

SonnyCampbellUnity (approved):

sindharta (approved):

mloccy (approved):
LGTM

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_preferences.py
seandillon92: > Does this check if the current hub version is 3.0.0?

No, msb_get_most_recent_version(a, b) returns the most recent version between a and b. So, if hub_version > 3.0.0 it will return the hub_version. If hub version < 3.0.0 it will return 3.0.0.  
 

> What happens if we're at a higher patch level?

If its 3.0.0 or a minor or major above it, it will return True. If its anything bellow 3.0.0 it will return False.

PR #299, opened 2022-11-15
blender: feedback for project creation 

seandillon92:


Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_preferences.py
```diff
@@ -282,4 +318,4 @@ def unregister():
     is_meshsync_in_manifest: bpy.props.BoolProperty(name = "Is Meshsync added in the package manifest", default = False)
     is_meshsync_in_manifest_lock: bpy.props.BoolProperty(name= "Is Meshsync resolved by the unity editor", default = False, update = on_in_package_lock_updated)
```
mloccy: Assuming this is user facing we probably want to capitalize Unity Editor

seandillon92: Good call, done!

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_preferences.py
```diff
@@ -167,33 +170,62 @@ def update_project_info(self):
...
-        if MESHSYNC_Preferences.thread is not None:
-            MESHSYNC_Preferences.thread.join(2.0)
+    def monitor_project_created(self, context):
+        while not self.cancel_create_project_thread:
+            sleep(0.5)
+            if msb_validate_project_path(self.project_path):
+                self.is_project_created = False
+                self.is_unity_project = True
+                self.redraw(context)
+                return
```
SonnyCampbellUnity, line 207: Does this need a timeout so it doesn't get caught forever if the project path check keeps failing?

SonnyCampbellUnity (commented):
Would also be useful to add a description/screenshot to the PR so it's clear what's expected to happen

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_preferences.py
seandillon92: I would be against that. The reason is we cannot have an accurate and meaningful timeout as each project takes minutes to create, and that can vary depending on the template, os resources, hardware etc.  This thread does not block the main thread and it will be terminated before a new thread with the same job is created. Also, Unity failing to create a project is a very unlikely scenario.

SonnyCampbellUnity (approved):

mloccy (approved):

PR #300, opened 2022-11-15
fix: Fix for crash when exporting baked transforms

schinkowski:


seandillon92 (approved):
Great stuff, would be good to have a root cause analysis at some point (doesn't have to be in this PR), so the team is aware of similar issues in the future. 
Also, potentially fixing https://github.com/unity3d-jp/MeshSync/issues/810 ?

sindharta (approved):

PR #302, opened 2022-11-16
blender-fix: send materials when server session changed.

schinkowski:


sindharta (approved):

schinkowski:
The client only sends textures when they have changed, which causes new servers to not receive textures if the client does not change them.
This change lets the client check whether the server has changed and if it did, reset the client's texture state so all textures are sent to the new server.

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
```diff
@@ -1326,7 +1326,22 @@ void msblenContext::logInfo(const char * format, ...)
...
+    // If server session changed, reset texture manager so textures are sent again:
+    if (m_sender.server_session_id != ms::InvalidID &&
+        server_session_id != m_sender.server_session_id) {
+        if (server_session_id != ms::InvalidID) {
+            resetMaterials();
+            // Ensure a full sync:
+            m_server_requested_sync = true;
+        }
+    }
+    m_sender.server_session_id = server_session_id;
```
seandillon92, line 1342: This is a polling function, is it possible to keep this as it is and move this code somewhere else? can we not use this in the sendObjects callback?

schinkowski: We don't really have another place to put it. I think this is the best place, the client finds out whether there is a server and what its session id is.

seandillon92 (approved):

sindharta (approved):

PR #309, opened 2022-11-21
blender-fix: avoid copies of linked duplicates

seandillon92:


sindharta (dismissed):

seandillon92:
Tested with an example file that includes nested properties in a geonode modifier, the solution works correctly.

PR #310, opened 2022-11-22
feat: DCC version analytics for Blender, 3dsMax, and Maya

schinkowski:


Plugins~/Src/MeshSyncClient3dsMax/msmaxContext.cpp
```diff
@@ -64,6 +66,33 @@ static void FeedDeferredCalls()
     ::PostMessage(GetCOREInterface()->GetMAXHWnd(), WM_TRIGGER_CALLBACK, (WPARAM)&FeedDeferredCallsImpl, (LPARAM)nullptr);
 }
 
+std::string getVersion()
+{
+    // 3dsmax is windows only so this should be fine:
+    FPValue resulted_value;
+    ExecuteMAXScriptScript(L"getFileVersion \"$max/3dsmax.exe\"",
```
SonnyCampbellUnity, line 73: is `ExecuteMAXScriptScript` a typo?

Plugins~/Src/MeshSyncClientBlender/msblenBinder.cpp
```diff
@@ -635,6 +635,23 @@ std::string abspath(const std::string& path)
...
+        py::eval<py::eval_mode::eval_statements>(
+            "import bpy\n"
+            "ret = bpy.app.version_string"
+            , global, local);
+        return (py::str)local["ret"];
+    }
+    catch (py::error_already_set& e) {
+        muLogError("%s\n", e.what());
+        return "";
+    }
```
seandillon92, line 652: There is a C API to get the version string https://github.com/blender/blender/blob/master/source/blender/blenkernel/BKE_blender_version.h  , have you tried using it?

SonnyCampbellUnity (dismissed):
One small comment about a typo?

Looks fine otherwise provided it's been tested on the various tools

Plugins~/Src/MeshSyncClient3dsMax/msmaxContext.cpp
schinkowski: Nope, that's their API https://help.autodesk.com/view/3DSMAX/2015/ENU/?guid=__files_GUID_5CC913A0_E7CC_457B_9EE4_EAA79B8AF2B4_htm

Plugins~/Src/MeshSyncClientBlender/msblenBinder.cpp
schinkowski: That would give us the version the plugin was compiled with. I want to get the version from the blender version currently running which may be slightly different.

seandillon92: Oh right, my bad

seandillon92 (dismissed):

sindharta (dismissed):

sindharta (approved):

seandillon92 (approved):

PR #319, opened 2022-12-06
WIP feat: Material baking

schinkowski:


Plugins~/Src/MeshSyncClientBlender/python/3.1.0/unity_mesh_sync.py
```diff
@@ -340,26 +339,29 @@ def draw(self, context):
...
-)
+    MESHSYNC_Preferences
+] + sharedClasses
 
 def register():
-    msb_initialize_properties()
     bpy.app.handlers.load_post.append(MESHSYNC_OT_AutoSync.load_handler)
     for c in classes:
         bpy.utils.register_class(c)
+    msb_initialize_properties()
```
seandillon92, line 353: why is this necessary? was it causing any issues because of the order?

Plugins~/Src/MeshSyncClientBlender/python/3.3.0/unity_mesh_sync.py
```diff
@@ -26,12 +27,6 @@
 mesh_sync_on_post_export = []
 
 
-class MESHSYNC_PT:
-    bl_space_type = "VIEW_3D"
-    bl_region_type = "UI"
-    bl_category = "Tool"
-
```
seandillon92: do we really need to remove this?

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_preferences.py
```diff
@@ -5,8 +5,8 @@
 
 from .unity_mesh_sync_installation import *
 from . import MeshSyncClientBlender as ms
-from bpy.types import AddonPreferences
-from bpy.props import StringProperty, IntProperty, BoolProperty
+from bpy.types import AddonPreferences, PropertyGroup
+from bpy.props import StringProperty, IntProperty, BoolProperty, IntVectorProperty, CollectionProperty
```
seandillon92, line 9: why do we import these new classes?

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_baking.py
```diff
@@ -0,0 +1,597 @@
...
+    bl_description = "Bakes textures, creates material copies and assigns the baked materials for all materials that " \
+                     "cannot be exported without baking them to textures"
+
+    def isMaterialCopy(self, mat):
+        return ORIGINAL_MATERIAL in mat
+
+    def canMaterialBeBaked(self, mat):
+        return mat is not None and\
+            not self.isMaterialCopy(mat) and\
+            mat.use_nodes
```
seandillon92, line 107: why do we not bake copies?

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_baking.py
```diff
@@ -0,0 +1,597 @@
...
+
+    def findMaterialOutputNode(self, mat):
+        node_tree = mat.node_tree
+
+        outputNode = None
+        for node in node_tree.nodes:
+            if node.type == 'OUTPUT_MATERIAL' and len(node.inputs[0].links) == 1:
+                outputNode = node
+                # Blender uses the last OUTPUT_MATERIAL node, so don't stop search here
```
seandillon92, line 116: What is the order of output nodes when traversing the tree? is it guaranteed to be as we expect it, or it could change in the future?

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_baking.py
```diff
@@ -0,0 +1,597 @@
...
+
+                bakedBSDF.location = (bsdf.location[0], minYLocation - 1000)
+                # Give bsdf a name and set its name on the material, so we can find it again:
+                bakedBSDF.name = BAKED_MATERIAL_SHADER
+                matCopy[BAKED_MATERIAL_SHADER] = bakedBSDF.name
+
+                matIndex = obj.material_slots.find(mat.name)
+                obj.material_slots[matIndex].material = matCopy
+            # obj.data.materials.pop(index=obj.data.materials.find(mat.name))
+            # obj.data.materials.append(matCopy)
```
seandillon92, line 436: do we need anything from this dead code? could we remove it?

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_baking.py
```diff
@@ -0,0 +1,597 @@
...
+        self.ensureUVs(context, obj)
+
+        scene = context.scene
+
+        if obj.mode == 'EDIT':
+            bpy.ops.object.mode_set(mode='OBJECT')
+
+        scene.render.engine = "CYCLES"
+        scene.cycles.device = "GPU"
+        scene.cycles.samples = 10       # TODO: This could be a setting for the user to change
```
seandillon92, line 449: I agree with the TODO comment. Also, does cycles support all bsdf that Eevee supports?

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_baking.py
```diff
@@ -0,0 +1,597 @@
...
+        scene.view_settings.view_transform = 'Standard'
+        scene.view_settings.look = 'None'
+        scene.view_settings.exposure = 0
+        scene.view_settings.gamma = 1
+        scene.sequencer_colorspace_settings.name = 'sRGB'
+        scene.cycles.time_limit = 0
+
+        # Set Cycles to use CUDA, if we aren't using any acceleration
+        if context.preferences.addons["cycles"].preferences.compute_device_type == "None":
+            context.preferences.addons["cycles"].preferences.compute_device_type = "CUDA"
```
seandillon92, line 466: Is this the only option available? What about OpenCL?

seandillon92: Also, would the Cuda option be ignored in non-Nvidia cards or will it throw an error?

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_baking.py
```diff
@@ -0,0 +1,597 @@
...
+        scene.cycles.use_auto_tile = True
+        scene.cycles.max_bounces = 12
+        scene.cycles.diffuse_bounces = 8
+        scene.cycles.glossy_bounces = 2
+        scene.cycles.transparent_max_bounces = 0
+        scene.cycles.transmission_bounces = 2
+        scene.cycles.volume_bounces = 0
```
seandillon92, line 473: Would it make sense to pass these values from the cycles panel?

seandillon92 (commented):
Looks good overall

Plugins~/Src/MeshSyncClientBlender/python/3.1.0/unity_mesh_sync.py
schinkowski: We are now using a property group and a collection properties of the new group, which requires the class for it to be registered before it can be used.

Plugins~/Src/MeshSyncClientBlender/python/3.3.0/unity_mesh_sync.py
schinkowski: I've just moved it to another file so it can be shared across different parts of the code more easily. This is the same for all versions and doesn't need a copy in each __init__ file.

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_preferences.py
schinkowski: I'm using them for baking

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_baking.py
schinkowski: This is to catch a scenario where a material copy we created is baked, which should never happen. That would bake textures that were already baked.

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_baking.py
schinkowski: From my testing this is what blender does. The node order in the nodetree can change but blender always uses the last material output node. It might change but there is no other way for us to detect which material output node is used, it's undefined behaviour and up to the user to set up correctly.

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_baking.py
schinkowski: As far as I know, cycles supports everything eevee supports and more.

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_baking.py
schinkowski: We're changing the settings on the cycles panel by setting those here.

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_baking.py
schinkowski: Need to check, maybe we can remove this entirely and let the user choose it.

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_preferences.py
seandillon92: But why are they imported in preferences? could they not be imported in the baking module?

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_baking.py
seandillon92: Just to make sure I follow:
I create a material in Blender.
I want to create a variation of the material so I create a copy of the first.
I modify the copy.
Will my copy be baked?

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_baking.py
seandillon92: oh right, is this necessary though? Maybe its better to let the user configure these?

seandillon92: also, if we decide to modify cycles settings through the script, should we revert them after the bake?

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_baking.py
schinkowski: If the copy was of an original, yes. If you copy a baked material, no. Modifications to baked materials cannot be supported unless we want to force the user to manually revert to the original material and just bake whatever is on the objects.

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_preferences.py
schinkowski: They could, but all the property imports used by other modules are in this file so I wanted to keep that consistent.

PR #320, opened 2022-12-06
blender-fix: export normals correctly in edit mode, taking polygon smooth into account

seandillon92:
In edit mode, we currently recalculate the normals we export, ignoring the smooth setting for each polygon. The fix is to lookup the corresponding polygon for each triangle, and based on the smooth flag, either use the vertex normal or the face normal.

This is fixing the issue with broken normals in edit mode, reported here https://github.com/unity3d-jp/MeshSync/issues/768

Demo of the fix:
ShadeModeChangeEditMode

### DIFF
--- a/Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
+++ b/Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
@@ -1122,10 +1122,10 @@ void msblenContext::doExtractEditMeshData(msblenContextState& state, BlenderSync
                 if (smooth) {
                     auto index =  idx->v->head.index;
                     auto vertext = vertices[index];
-                    dst.normals[ii++] = to_float3(vertext->no);
+                    dst.normals[ii++] = ms::ceilToDecimals(to_float3(vertext->no));
                 }
                 else {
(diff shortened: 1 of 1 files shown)


sindharta (dismissed):

schinkowski (dismissed):

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
```diff
@@ -1112,8 +1113,21 @@ void msblenContext::doExtractEditMeshData(msblenContextState& state, BlenderSync
...
+            const int polygon_index = triangle[0]->f->head.index;
+
+            auto polygon = polygons[polygon_index];
+            auto smooth = polygon->head.hflag & BM_ELEM_SMOOTH;
+
+            for (struct BMLoop* idx : triangle) {
+                if (smooth) {
+                    auto index =  idx->v->head.index;
+                    auto vertext = vertices[index];
+                    dst.normals[ii++] = to_float3(vertext->no);
```
seandillon92: Just realized, might be good to use the decimal rounding function to avoid issues where the object is sent across on every frame.

schinkowski (approved):

PR #327, opened 2022-12-12
blender-fix: Fix for crashes when instantiating curves through geonodes

seandillon92:
This PR cleans some tech debt we have in geonodes. With the recent work done to support any object and not just Meshes, a bug from an pre-existing hack I had introduced has come to surface. Specifically this line:
`std::memcpy(&rec.object_copy, obj, sizeof(Object));`
does a bitwise copy on the Object that is instantiated. There are 2 issues with this:
1. This operation is unsafe. obj could be using padding bytes
2. We copy pointers which may end up pointing to released or changed memory.

While this has worked for Meshes so far, I started noticing crashes when synching files with curves.

The fix is to refactor the code and instead of copying the object that is references, export it immediately.

### DIFF
--- a/Plugins~/Src/MeshSyncClientBlender/msblenContextGeometryNodes.cpp
+++ b/Plugins~/Src/MeshSyncClientBlender/msblenContextGeometryNodes.cpp
@@ -8,13 +8,13 @@
 ms::InstanceInfoPtr msblenContext::exportInstanceInfo(
     msblenContextState& state,
     msblenContextPathProvider& paths,
-    Object* instancedObject,
-    Object* parent,
+    ms::TransformPtr transform,
+    ms::TransformPtr parent,
     SharedVector<mu::float4x4> mat) {
 
     auto info = ms::InstanceInfo::create();
-    info->path = paths.get_path(instancedObject);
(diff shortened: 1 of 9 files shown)


Plugins~/Src/MeshSyncClientBlender/msblenContextGeometryNodes.cpp
```diff
@@ -27,57 +27,73 @@ void msblenContext::exportInstances() {
...
+                settings.multithreaded = false;
+
+                auto transform = exportObject(*m_instances_state, m_intermediate_paths, settings, rec.obj, false, true);
+                transform->reset();
+                exportedTransforms[rec.id] = transform;
+            }
+            else if (scene_objects.find(rec.name) == scene_objects.end()) {
+                auto settings = m_settings;
+                settings.multithreaded = false;
+                settings.BakeModifiers = false;
```
seandillon92, line 60: I did some testing and it turns out that when we access objects from another scene, we get an evaluated object, i.e the modifiers are already baked.

sindharta-tanuwijaya (approved):

PR #329, opened 2023-01-06
feat: Material baking

schinkowski:


Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_common.py
```diff
@@ -199,14 +200,18 @@ def meshsync_prepare():
...
-    for f in mesh_sync_on_pre_export:
-        f()
-
-def meshsync_post_export():
-    for f in mesh_sync_on_post_export:
-        f()
-
-def meshsync_server_requested_callback():
-    for f in mesh_sync_server_requested_callback:
-        f()
```
seandillon92: have these been moved somewhere?

Documentation~/Blender.md
```diff
@@ -94,3 +95,32 @@ The following is a list of supported nodes and how they are handled.
...
+If the object has no UVs, MeshSync will use blender's *Smart UV Project* operator to generate UVs.
+
+The blender console will show progress during the bake.
+
+## Material baking troubleshooting
+|**Problem** |**Possible cause** |
+|:---     |:---|
+| The baked maps are black | There are many possible causes. Is the object UV unwrapped? Baking will not work without valid UVs. Metallic materials do not bake well in blender. |
+| The baked maps do not match the original material | The material node tree does not use a BSDF as input to the *Material Output* node and the lower quality fallback bake mode had to be used. |
+| Geometry node instances are missing | To bake materials, all modifiers need to be applied. When geometry node modifiers are applied, instances are lost. Use a *Realize Instances* node to make the instances real geometry. |
```
seandillon92: Does the baking script apply all the modifiers? is that reversed after the baking?

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_baking.py
```diff
@@ -0,0 +1,1195 @@
...
+    VERBOSE = 0
+    NORMAL = 1
+    ERROR = 2
+
+
+# For debugging and getting a callstack on error:
+throwExceptions = True
+showLogLevel = LogLevel.VERBOSE
+
+# Commented out ones will be supported in next version:
```
SonnyCampbellUnity: Are any of these meant to be commented out?

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_baking.py
```diff
@@ -0,0 +1,1195 @@
...
+    def findMaterialOutputNode(self, node_tree):
+        outputNode = None
+        for node in node_tree.nodes:
+            if node.type == 'GROUP':
+                outputNodeInGroup = self.findMaterialOutputNode(node.node_tree)
+                if outputNodeInGroup is not None:
+                    outputNode = outputNodeInGroup
+            elif node.type == 'OUTPUT_MATERIAL' and len(node.inputs[0].links) == 1:
+                outputNode = node
+                # Blender uses the last OUTPUT_MATERIAL node, so don't stop search here
```
SonnyCampbellUnity: If you have two Material Output nodes and switch which one you're using by changing the link, will it definitely re-order the list to set the Material Output node you're using as the last Material Output node?

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_baking.py
```diff
@@ -0,0 +1,1195 @@
...
+                obj.select_set(False)
+                obj.hide_set(wasHidden)
+
+                if bakeSettings.run_modal:
+                    yield
+                    context = self.context
+
+            if UV_OVERRIDE in obj.data and len(obj.data.uv_layers) > 1:
+                msb_log(
+                    f"New UVs were generated for '{obj.name}' for baking. Old UVs need to be deleted so the baked textures work correctly.")
```
SonnyCampbellUnity: Could this delete a user's UVs accidentally? Can the user specify if they want their UVs overridden?

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_baking.py
```diff
@@ -0,0 +1,1195 @@
...
+
+    def getNodeYLocation(self, node):
+        location = node.location[1]
+        if node.parent is not None:
+            location += self.getNodeYLocation(node.parent)
+        return location
+
+    def prepareMaterial(self, context, obj, bsdf, mat, canBakeBSDF):
+        # Make material copy if this is not a copy already:
+        if ORIGINAL_MATERIAL not in mat:
```
SonnyCampbellUnity: Can you invert these if statements to reduce the nesting?

```
if ORIGINAL_MATERIAL in mat:
   return mat

// Do logic stuff 
```

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_baking.py
```diff
@@ -0,0 +1,1195 @@
...
+            # Use existing copy if there is one:
+            matCopyName = f"{mat.name}_{obj.name}_baked"
+
+            matCopyIndex = bpy.data.materials.find(matCopyName)
+            if matCopyIndex >= 0:
+                matCopy = bpy.data.materials[matCopyIndex]
+
+                # Replace material with its baking copy:
+                matIndex = obj.material_slots.find(mat.name)
+                obj.material_slots[matIndex].material = matCopy
```
SonnyCampbellUnity: Same here

```
if matCopyIndex >= 0:
    matCopy = bpy.data.materials[matCopyIndex]

    # Replace material with its baking copy:
    matIndex = obj.material_slots.find(mat.name)
    obj.material_slots[matIndex].material = matCopy
    mat = matCopy
    return mat

// Do else clause logic 
```

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_baking.py
```diff
@@ -0,0 +1,1195 @@
...
+        if obj.mode == 'EDIT':
+            bpy.ops.object.mode_set(mode='OBJECT')
+
+    def getNodeYLocation(self, node):
+        location = node.location[1]
+        if node.parent is not None:
+            location += self.getNodeYLocation(node.parent)
+        return location
+
+    def prepareMaterial(self, context, obj, bsdf, mat, canBakeBSDF):
```
SonnyCampbellUnity, line 775: Can you add a comment with this method. I'm not fully sure what it's doing

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_baking.py
```diff
@@ -0,0 +1,1195 @@
...
+        self.prepareObjectForBaking(context, obj)
+        mat = self.prepareMaterial(context, obj, bsdf, mat, canBakeBSDF)
+
+        return mat
+
+    def setupRenderSettings(self, context):
+        self.originalSceneSettings = []
+
+        self.setRestorableContextSetting(context, "scene.render.engine", "CYCLES")
+        self.setRestorableContextSetting(context, "scene.cycles.device", "GPU")
```
SonnyCampbellUnity: Do we need to enforce CYCLES/GPU? Could there be a scenario where that doesn't work?

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_baking.py
```diff
@@ -0,0 +1,1195 @@
...
+        context.window_manager.modal_handler_add(self)
+
+        # Store area for later use, these won't exist on modal timer callbacks:
+        self.area = context.area
+
+        return self.execute(context)
+
+    def modal(self, context, event):
+        # Allow cancellation by pressing escape:
+        if event.type == 'ESC':
```
SonnyCampbellUnity, line 1142: Is it documented or shown to the user that they can cancel the process by hitting ESC?

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_baking.py
```diff
@@ -0,0 +1,1195 @@
...
+            self.incrementProgress(context, "Baking cancelled by user", reset=True)
+            return {'FINISHED'}
+
+        # Refresh context each run:
+        self.context = context
+        for _ in self.bakeTask:
+            return {'RUNNING_MODAL'}
+
+        msb_log(f"Finished baking. Time taken: {datetime.timedelta(seconds=(time.time() - self.startTime))}",
+                LogLevel.ERROR)
```
SonnyCampbellUnity, line 1153: Is this supposed to be an error?

SonnyCampbellUnity (commented):
Added some questions and comments

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_baking.py
schinkowski: Yes, UVs can be lost or overwritten. The user can choose whether to generate UVs and risk that or not.

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_baking.py
schinkowski: Done

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_baking.py
schinkowski: Done

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_baking.py
schinkowski: Done

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_baking.py
schinkowski: We need cycles for baking, GPU is faster.
I've added a check now to not do this if there is no GPU.

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_baking.py
schinkowski: I've put it in the docs

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_baking.py
schinkowski: It's the highest log level, to ensure the message always shows.

Documentation~/Blender.md
schinkowski: It's optional. We can't revert that because that would not send the correct mesh when it's synced after baking.
We could do backups of any modified objects and restore those but it would slow down the bake and increase memory usage, so I think it's better to let the user do their own backups if needed.

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_baking.py
schinkowski: Not anymore. I've removed that comment now.

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_baking.py
schinkowski: I've improved that part of the code now. There's a better way to check that.

Documentation~/Blender.md
seandillon92: can we not just bake on the evaluated objects?

schinkowski: No, the bake is a blender operation that works on the scene objects.

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_baking.py
SonnyCampbellUnity: Does the user know their UVs will be lost/overwritten?

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_baking.py
SonnyCampbellUnity: Can you explain a little context about why it is creating a copy and what is involved...

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_baking.py
SonnyCampbellUnity: Is there anywhere we can show this on the UI? Something like "In Progress (Press ESC to Cancel)"? Users will often skim docs or ignore them entirely and might miss it otherwise

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_baking.py
SonnyCampbellUnity: Does it say "ERROR" in the message though? Or does it just show up red or something?

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_baking.py
schinkowski: It's in the docs and they can turn UV generation on or off, so they should be aware.

Documentation~/Blender.md
```diff
@@ -94,3 +95,36 @@ The following is a list of supported nodes and how they are handled.
...
+## Material baking
+
+The basic material sync mode only supports bsdf inputs that are textures or constant values.
+MeshSync has functionality to create baked material copies for each object to allow syncing of procedural materials.
+
+|**Option** |**Description** |
+|:---     |:---|
+| Objects to bake | Bakes all objects in the scene or only selected objects (including hidden objects). |
+| Material channels to bake | Choose which channels should be baked. |
+| Generate UVs | Off: Uses UVs on the object, If needed: Generates UVs if there are no usable UVs on the object, Always: Always generates UVs for baking, even if there are existing UVs. NOTE: These settings can be destructive to existing UVs. |
```
seandillon92: Could "Off", "Always" and "Note" start their own line? I think it might a bit easier to read that way

Documentation~/Blender.md
```diff
@@ -94,3 +95,36 @@ The following is a list of supported nodes and how they are handled.
...
+| Apply modifiers | Applies all modifiers on the object to ensure the UV coordinates are correct for baking. |
```
seandillon92: Should we warn that this is not reversible?

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_baking.py
```diff
@@ -0,0 +1,1202 @@
...
+}
+
+
+def msb_log(text, level=LogLevel.VERBOSE):
+    if level >= showLogLevel:
+        print(text)
+
+
+def msb_canObjectMaterialsBeBaked(obj: bpy.types.Object) -> bool:
+    hasMaterials = obj.data is not None and obj.type == 'MESH' # hasattr(obj.data, 'materials')
```
seandillon92: can the # hasattr(obj.data, 'materials') be removed?

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_baking.py
```diff
@@ -0,0 +1,1202 @@
...
+    return True
+
+
+# Methods to help getting and setting nested attributes:
+def msb_rsetattr(obj, attr, val):
+    pre, _, post = attr.rpartition('.')
+    return setattr(msb_rgetattr(obj, pre) if pre else obj, post, val)
+
+
+def msb_rgetattr(obj, attr, *args):
```
seandillon92, line 65: what does the r in msb_rgetattr stand for? recursive?

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_baking.py
```diff
@@ -0,0 +1,1202 @@
...
+                                         description="Whether to auto-generate UVs",
+                                         items=(('OFF', 'Off',
+                                                 'Bake all objects in the scene (including hidden)'),
+                                                ('IF_NEEDED', 'If needed',
+                                                 'Automatically UV unwraps objects if there are no UVs or existing UVs are not in the 0..1 range. WARNING: This will delete existing UVs on the object!'),
+                                                ('ALWAYS', 'Always',
+                                                 'Always automatically UV unwraps objects. WARNING: This will delete existing UVs on the object!')),
+                                         default='OFF')
+    apply_modifiers: bpy.props.BoolProperty(name="Apply modifiers",
+                                            description="In order to bake and get correct UVs, all modifiers need to be applied",
```
seandillon92: Should we add a warning here as well, similar to the one about UVs?

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_baking.py
```diff
@@ -0,0 +1,1202 @@
...
+
+
+def msb_bakeAllChanged(self, context):
+    bakeAll = context.scene.meshsync_bake_settings.bake_all_channels
+    for channelSetting in context.scene.meshsync_bake_settings.bake_channel_settings:
+        channelSetting.bakeChannelEnabled = bakeAll
+
+
+@persistent
+def msb_setBakingDefaults(dummy):
```
seandillon92, line 140: why do we need the dummy argument?

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_baking.py
```diff
@@ -0,0 +1,1202 @@
...
+        materials = []
+        finalMaterials = []
+        for matSlot in obj.material_slots:
+            materials.append(matSlot.material)
+
+        if bakeIndividualMats:
+            for matSlot in obj.material_slots:
+                matSlot.material = None
+
+        try:
```
seandillon92, line 458: this is a long try except clause. We could extract everything under Try to a bakeMaterials function, to make it a bit easier to read

Documentation~/Blender.md
schinkowski: It's hard with this table markdown language, they would look like separate options instead of items in the enum

Documentation~/Blender.md
schinkowski: I've added it.

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_baking.py
schinkowski: We need to keep the original material so the bake can be reverted. 
The material copy is modified to bake the individual channels by connecting them directly to the material output and baking it as emission. After all the baking is complete, all nodes in the material copy, except the BSDF and material output and baked images are deleted so the material only uses the baked images.

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_baking.py
```diff
@@ -0,0 +1,1202 @@
...
+            for matIndex, mat in enumerate(finalMaterials):
+                obj.material_slots[matIndex].material = mat
+
+    def bake(self):
+        context = self.context
+
+        # Make sure meshsync is finished and ready:
+        from .unity_mesh_sync_common import msb_apply_scene_settings, msb_context
+
+        # if not msb_context.is_setup:
```
seandillon92: dead code?

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_baking.py
schinkowski: The log level just defines whether the logs should show or not. It's set at the top of the file. For releases it should be set to ERROR, for debugging VERBOSE.

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_baking.py
```diff
@@ -0,0 +1,1202 @@
...
+            for col in context.scene.collection.children_recursive:
+                if col.name == hiddenCollection:
+                    col.hide_render = True
+
+    def checkForUV0(self, obj, uvMapName, channel):
+        '''
+        :param obj: Object the material is on
+        :param uvMapName: Name of the UV map
+        :param channel: Input channel to process
+        :return: [True, Reason] if the given UV map name is not UV0 of the given object. Otherwise [False].
```
seandillon92, line 549: The name is a bit confusing, would checkIsNotUV0 describe the function more accurately?

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_baking.py
schinkowski: I've added it now.

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_baking.py
```diff
@@ -0,0 +1,1204 @@
...
+                msb_log(
+                    f"Cannot bake '{obj.name}' because it does not have the UV map {uvMapName} that the input for '{channel}' needs.")
+                return [False]
+
+            if obj.data.uv_layers[0].name == uvMapName:
+                return [False]
+
+            return [True, "Image UV input is not UV0."]
+
+    def doesImageNodeUseUv0(self, obj, link, imageNode, channel):
```
seandillon92: similarly here, from the name I expect this to return True when UV0 is used

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_baking.py
```diff
@@ -0,0 +1,1204 @@
...
+            msb_log(f"Baking {channel} for '{obj.name}'. Reason: {result[1]}")
+            bakedMat = self.bakeChannel(context, obj, mat, bsdf, matOutput, channel)
+            return True, bakedMat
+
+        return False, mat
+
+    def createImage(self, context, obj, name, colorSpace, alpha=False):
+        imageName = name.replace(" ", "_")
+
+        # Delete any existing image with this name to ensure the dimensions and alpha settings are correct:
```
seandillon92, line 667: Is it guaranteed that it won't delete any image the user has created?

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_baking.py
schinkowski: Yeah

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_baking.py
schinkowski: Yeah, done

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_baking.py
schinkowski: Yeah, done

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_baking.py
schinkowski: This method is called from blender's load_post event, which requires the method to have 1 parameter. We don't use it.

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_baking.py
schinkowski: Done

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_baking.py
schinkowski: Removed

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_baking.py
schinkowski: I've renamed it

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_baking.py
schinkowski: I renamed it

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_baking.py
schinkowski: In theory we may delete user's images. The probability that they named the image in the blend file exactly the same as the name we generate is pretty low. We need the name to be deterministic to ensure we delete previously baked images before baking.

seandillon92: is it possible to have a pattern when naming the images? for example `ms_baker_<map type>.png` or similar?

seandillon92: I am not sure if currently we use names such as "albedo.png". What is the naming convention?

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_baking.py
SonnyCampbellUnity: I meant could you add the explanation to the comment, so it's clear for anyone who needs to understand the function in future.

schinkowski: Done

SonnyCampbellUnity (approved):

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_baking.py
schinkowski: Images are named like this: `<original_material_name>_<object_name>_baked_<channel_name>`

seandillon92: Sounds good, that should be specific enough

seandillon92 (approved):

PR #330, opened 2023-01-06
fix: materials reference and roughness fix

schinkowski:


SonnyCampbellUnity (approved):

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
```diff
@@ -853,9 +853,20 @@ void msblenContext::doExtractNonEditMeshData(msblenContextState& state, BlenderS
     const size_t num_polygons = polygons.size();
     size_t num_vertices = vertices.size();
 
-    std::vector<int> mid_table(mesh.totcol);
-    for (int mi = 0; mi < mesh.totcol; ++mi)
-        mid_table[mi] = getMaterialID(mesh.mat[mi]);
+    std::vector<int> mid_table(bobj.m_ptr->totcol);
```
seandillon92, line 856: is bobj.m_ptr-totcol guaranteed to be the same as mesh.totcol?

seandillon92 (approved):

Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
schinkowski: Yeah, this is what the blender code uses.

PR #342, opened 2023-01-19
blender: toggle for realizing instances when baking

seandillon92:


Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_baking.py
```diff
@@ -369,6 +373,13 @@ def preBakeObject(self, obj):
                 # Can't apply modifiers with shared data:
                 bpy.ops.object.make_single_user(type='SELECTED_OBJECTS', obdata=True)
                 for mod in obj.modifiers[:]:
+                    if bakeSettings.realize_instances and mod.type == "NODES":
+                        output = mod.node_group.nodes["Group Output"]
```
schinkowski: That name is not guaranteed to exist, users can rename nodes. I think you'll need to loop through the nodes and find nodes with the output type and probably check which is the active output in case there are more than 1. I'm doing something similar in `findMaterialOutputNode` for the material graph.

seandillon92: I changed it to use the type. I could not find a reliable way to get the right group output, since active would point to the last selected output node or -1 if none has been selected yet. So instead, I just iterate over all output nodes and add the realize instances, assuming it will be a small number.

Plugins~/Src/MeshSyncClientBlender/python/unity_mesh_sync_baking.py
```diff
@@ -347,6 +351,19 @@ def incrementProgress(self, context, message, obj=None, reset=False):
 
         bakeSettings.bake_message = message
 
+    def addRealizeInstances(self, mod):
+        nodes = mod.node_group.nodes
+        outputs = [x for x in nodes if x.type == "GROUP_OUTPUT"]
+
+        for output in outputs:
+            if len(output.inputs) == 0 or len(output.inputs[0].links) == 0:
```
schinkowski, line 359: I think instead of checking the number of inputs, the 'is_active_output' flag should be used.

seandillon92: I tried that but noticed that is_active_input is true for the node that is selected last. So if have 2 outputs nodes and only one has links but I select the empty one, the bake will not include the instances. 

My solution is to simply try and add the realize instance node if possible, that is the output node has at least one input with at least one link.

schinkowski: Yeah the selected node should be the active one, that's how it works for materials as well. Would there be any harm to just add the realize instances node before every output?

PR #348, opened 2023-01-25
fix: Mix shader fix

schinkowski:


Plugins~/Src/MeshSyncClientBlender/msblenMaterialsExportHelper.cpp
```diff
@@ -68,16 +78,38 @@ bNode* handleBSDFTypes(const Material* mat, bNode* bsdf) {
...
+		// For mix shaders, prefer output based on fraction if it's connected:
+		if (bsdf->type == SH_NODE_MIX_SHADER)
+		{
+			auto factorSocket = getInputSocket(bsdf, "Fac");
+			if (factorSocket)
+			{
+				// Can only evaluate default value on the socket:
+				if (!traverseReroutes(getNodeConnectedToSocket(factorSocket), mat)) {
+					auto factor = (bNodeSocketValueFloat*)factorSocket->default_value;
+					if (factor->value > 0.5f) {
```
seandillon92, line 103: I think this needs to be inverted

seandillon92 (dismissed):

PR #360, opened 2023-02-14
[DONOTMERGEYET] poco install instruction

seandillon92:
Some clarification on the Poco_DIR instruction. A few people in my team were confused by this.

### DIFF
--- a/Plugins~/Docs/en/BuildDCCPlugins.md
+++ b/Plugins~/Docs/en/BuildDCCPlugins.md
@@ -33,7 +33,9 @@
     
     > To build Poco libraries with other configurations, see [Poco's Getting Started](https://pocoproject.org/docs/00200-GettingStarted.html).
     
-1. Add *Poco_DIR* environment variable to point to the Poco root folder above.
+1. Add *Poco_DIR* environment variable to point to the Poco root folder above. 
+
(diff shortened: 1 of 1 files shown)


Plugins~/Docs/en/BuildDCCPlugins.md
```diff
@@ -33,7 +33,9 @@
     
     > To build Poco libraries with other configurations, see [Poco's Getting Started](https://pocoproject.org/docs/00200-GettingStarted.html).
     
-1. Add *Poco_DIR* environment variable to point to the Poco root folder above.
+1. Add *Poco_DIR* environment variable to point to the Poco root folder above. 
+
+For example, if you poco installation is located in C:\Users\Username\Downloads\poco-poco-1.10.1-release\, the environment variable should point to C:\Users\Username\Downloads\poco-poco-1.10.1-release\poco-poco-1.10.1-release\
```
sindharta, line 38: This may actually confuse more. Some unzipping tools may create an additional folder when extracting (depending on the config too), and in this case, we do need to point to the lower level folder.
But aside from this case, this example is actually not correct.

My `Poco_DIR` for example, points to `E:\SDK\poco`.

I think we should use a term that describes:
* a folder where POCO files really exist, 
* instead of a folder that contains another folder containing those files.

seandillon92:
Closing for now, will reopen when I have more time to work on this

PR #365, opened 2023-03-21
feat: Adding option to only sync selected objects.

schinkowski:


Plugins~/Src/MeshSyncClientBlender/python/3.3.0/unity_mesh_sync.py
```diff
@@ -335,7 +336,6 @@ def draw(self, context):
     MESHSYNC_PT_Animation,
     MESHSYNC_PT_Cache,
     MESHSYNC_PT_Version,
-    MESHSYNC_OT_SendObjects,
```
seandillon92, line 338: Is the current manual sync button replaced by the selected only manual sync? I wonder if it would be better to have a toggle on the sync settings named "Selected Only".

schinkowski: The toggle would then also need for auto-sync. We could add that later if needed.

schinkowski: The button is not replaced, I'm adding a second button.

seandillon92: > The toggle would then also need for auto-sync. We could add that later if needed.

Yeah I think it should apply for both modes.

> The button is not replaced, I'm adding a second button.

Ok cool, as long as we don't remove the current functionality, it sounds good to me.

seandillon92 (approved):

Plugins~/Src/MeshSyncClientBlender/python/3.3.0/unity_mesh_sync.py
schinkowski: For auto-sync it's not that easy to only send the selected object because changing one object could affect another via drivers, modifiers, etc. 
Also, auto-sync only sends updated objects, changing the selection would not send the new object without doing a lot of extra work to make it do that. I don't really need auto-sync for selected objects and just wanted manual sync.
Considering all that, I think we can do auto-sync when users ask for it.

seandillon92: Sounds good, if we are not replacing the existing button, there is no need for that toggle right now anyway.

sindharta (approved):

PR #376, opened 2023-05-23
blender-fix: null reference check in visible_in_collection()

schinkowski:


Plugins~/Src/MeshSyncClientBlender/msblenUtils.cpp
```diff
@@ -89,16 +89,18 @@ bool visible_in_viewport(const Object *obj)
 
 bool visible_in_collection(LayerCollection* lc, const Object* obj) {
     // Check if the object is in the layer collection, if it is, check if the layer is excluded:
-    for (auto collectionObject : blender::list_range((CollectionObject*)lc->collection->gobject.first)) {
-        if (collectionObject->ob == obj) {
-            if ((!(lc->flag & LAYER_COLLECTION_EXCLUDE)) &&
+    if (lc->collection) {
```
sindharta, line 92: I think we should just use early return here to reduce code diffs  (and code indentation for the code that really matters.)

```
if (!lc->collection)
    return
```

schinkowski: There's another check below that needs to be done, we can't exit early

PR #377, opened 2023-05-23
fix: apply mirror Modifier for instances when applicable

schinkowski:


Plugins~/Src/MeshSyncClientBlender/msblenContext.cpp
```diff
@@ -802,7 +802,7 @@ void msblenContext::doExtractMeshData(msblenContextState& state, BlenderSyncSett
             doExtractNonEditMeshData(state, settings, dst, obj, data);
         }
 
-        if (!settings.BakeModifiers&& !is_editing) {
+        if (!settings.BakeModifiers && !is_editing && state.manager.needsMirrorBaking()) {
```
sindharta: This code seems to be run when NOT baking, so the name `needsMirrorBaking` seems to be misleading ?

schinkowski: We're manually baking the mirror modifier when baked modifiers is turned off. It does still bake the modifier, just not with blender.

schinkowski: I renamed it now.

PR #378, opened 2023-05-23
[For 0.18.x] fix: Instancing fixes

schinkowski:
We had an architectural problem with the way we send instances, which this PR addresses.
Previously we sent one instance info for each instance, which caused problems when the same object is instanced on multiple parents.
With these changes each instance has a collection of parents it appears on, which allows us to parent the instances to the correct parent and allows us to do nested instances correctly.


sindharta:
Same as for the MeshSync side, can you please elaborate what kind of particular bug that this PR fixes ?
Also, does this depends on https://github.com/unity3d-jp/MeshSync/pull/921 ?

seandillon92:
Could you create a new PR against blender-3.5-support2 so we can see only the relevant changes?

schinkowski:
> Could you create a new PR against blender-3.5-support2 so we can see only the relevant changes?

I did it here: https://github.com/Unity-Technologies/MeshSyncDCCPlugins/pull/379/files

PR #380, opened 2023-06-02
fix: use path provider to ensure instances use their correct path for properties

schinkowski:


seandillon92 (approved):
LGTM, in what scenario do you need to export properties from a non-default path?

sindharta (approved):

PR #382, opened 2023-06-06
[For 0.18.x] blender-change: don't use normals to calculate hash in Blender

schinkowski:


sindharta:
Related to:
https://github.com/unity3d-jp/MeshSync/pull/920#issuecomment-1578343865

sindharta (approved):

